PR #4, opened 2014-06-10
Allow to return immediatly from queue.take() if there are no ready tasks

rybakit:
```
queue.take(0, 'q')     -- waits indefinitely
queue.take(0, 'q', -1) -- waits indefinitely
queue.take(0, 'q', 0)  -- returns immediately
queue.take(0, 'q', 1)  -- waits 1 second and returns
```


kostja:
Hi, would you still like this request to be pushed?

rybakit:
Yes, I would like that.

PR #15, opened 2015-03-13
Added new queue `stube` which based on `utube`

zloidemon:
* Added support pass opts to :take()
* Added info about stube in readme
* And source of queue with tests


Sulverus:
Merged with master in stube branch(https://github.com/tarantool/queue/tree/stube)

PR #23, opened 2015-06-26
Simplify stats resulting table

rybakit:


Sulverus:
After fixes in tarantool, this version of queue.statistics() can be called using tarantool-protocol(e.g net.box.call). Please update tarantool to the latest version

rybakit:
Thanks.

PR #48, opened 2016-12-16
Support touch function

### LINKED ISSUE #47: FR: add method "prolong ttr"
Hello.

We are using queue for passing tasks to background service. Unfortunately, at the moment of task creation exact volume of work is unknown and can vary from task to task on orders of magnitude. I would like to set low TTR on all tasks initially in such way that a majority of tasks will succeed before TTR will expire

But other, larger tasks need to prolong TTR by somehow. And it would be great if this will be done by queue interface, not by some hacks via underlying spaces.

Would you accept this feature and should I start to work on pull request?

aensidhe:
Closes #47

---
This change is [Reviewable](https://reviewable.io/reviews/tarantool/queue/48)

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,7 +385,9 @@ time needed to work on task.
 
 Effect: the value of `ttr` and `ttl` increased by `increment` seconds. If queue
 does not support ttr, error will be thrown. If `increment` is lower than zero,
-error will be thrown.
+error will be thrown. If `increment` is zero effect is no op. If `increment` is
+`nil` it will be equal to current `ttr` of task. If current `ttr` of task is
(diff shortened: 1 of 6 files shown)


aensidhe:
TODO:
- [] rename extend_ttr to touch
- [] if we're extending ttr - extend ttl to same interval
- [] if ttr or ttl larger than TIMEOUT_INFINITY, then touch is noop.

aensidhe:
Closed in https://github.com/tarantool/queue/commit/b2016c7f28940adf35ea34f8185659e5c4a17475

rybakit:
@bigbes @aensidhe What about documenting the new method?

PR #75, opened 2018-02-16
fix: avoid calling on_task_change with nil

### LINKED ISSUE #74: why call on_task_change with nil?
https://github.com/tarantool/queue/blob/593c9f6f949c84865bcd383d542876ebbc93520a/queue/abstract/driver/utube.lua#L102-L104

We shouldn't call `on_task_change(nil, 'delete')` since a nil task indicates the task has already been removed from the queue, which means `on_task_change` was already called before when the task was previously deleted or acknowledged.

I assume this is being done in other queue types, too.

I'll send a pull request if you agree.

aleclarson:
This fixes #74 

Correct me if I'm wrong, but when `self.space:get(id)` returns nil, that means the task has already been deleted or acknowledged, so there's no point in calling `on_task_change(nil, 'delete')`.

### DIFF
--- a/queue/abstract/driver/utube.lua
+++ b/queue/abstract/driver/utube.lua
@@ -93,14 +93,12 @@ function method.delete(self, id)
     self.space:delete(id)
     if task ~= nil then
         task = task:transform(2, 1, state.DONE)
+        self.on_task_change(task, 'delete')
 
         local neighbour = self.space.index.utube:min{state.READY, task[3]}
-        self.on_task_change(task, 'delete')
         if neighbour then
             self.on_task_change(neighbour)
         end
-    else
-        self.on_task_change(task, 'delete')
     end
     return task
 end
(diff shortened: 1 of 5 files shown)


queue/abstract/driver/utube.lua
```diff
@@ -93,14 +93,12 @@ function method.delete(self, id)
     self.space:delete(id)
     if task ~= nil then
         task = task:transform(2, 1, state.DONE)
+        self.on_task_change(task, 'delete')
```
LeonidVas: What is the purpose of the on_task_change() moving?

aleclarson: I don't remember. You can put it back below the `local neighbor` declaration to be safe.

PS: I'm not using tarantool atm, so you'll have to wrap this PR up on your own. ;)

LeonidVas: Ok

LeonidVas:
> I think test cases should be added to show the changed behaviour of a user-provided `on_task_change` callback.

The behavior will not change. Before the deletion in abstract.lua, it will call the peek and throws an error on failure.

LeonidVas:
Patches has been pushed to master.

PR #76, opened 2018-02-16
feat: add `concurrent` option to utube

aleclarson:
Inspired by #15, this commit adds a `concurrent: number` option to the `utube` queue type.

Additionally, you can now do `:take({ utube = '...' })` to take from a specific sub-queue.

The default behavior of `utube` has not changed.
Only one task per `utube` sub-queue may run at any time.

Set `concurrent` to `2` and only two tasks per sub-queue may run at any time.
Set `concurrent` to `1/0` or `math.huge` and no throttling is enforced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,14 @@ options are equal.
 
 The main idea of this queue backend is the same as in a `fifo` queue:
 the tasks are executed in FIFO order.
-However, tasks may be grouped into sub-queues.
+
+However, tasks may be grouped into sub-queues. Sub-queues split the task
+stream according to the sub-queue name. Each sub-queue is executed in strict
+FIFO order. By default, only one task per sub-queue may be taken at any time.
 
 The following options can be specified when creating a `utube` queue:
+  * `concurrent` - number - the number of concurrent tasks per sub-queue
+(defaults to 1)
   * `temporary` - boolean - if true, the contents of the queue do not persist
 on disk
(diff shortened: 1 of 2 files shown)


aleclarson:
Alright, ready for a code review.

The `is_throttled` method uses the most efficient means of determining concurrency for the given sub-queue name.

When `:take` is called without a specific sub-queue name, we traverse all `READY` tasks, but make sure to avoid calling `is_throttled` for the same sub-queue more than once. I removed the `GE` iterator, because it didn't look like it was necessary. Correct me if I'm wrong there.

LeonidVas:
Hi. Thank you for the patch. Sorry for the big latency. Please, could  you to describe few use case of the feature?

aleclarson:
It's been a while, but I guess being able to `:take` multiple tasks from a sub-queue at one time had a good enough use case for me to implement it. :) 

Is there another way to achieve that already?

LeonidVas:
As work around: "You can use some manager around fifo/fifottl".
I agree that this may be a good ability, but without understanding the purpose of the ability, it is useless.

aleclarson:
To me, the purpose is obviously useful. There are times where you want to process multiple tasks from the same sub-queue at once. You can't do that without this PR. Not sure what more could be explained here. ¯\\_(ツ)_/¯

LeonidVas:
After some discussion, we decided not add this feature to the queue. If the situation change, we get back to the PR.

Issue #80, opened 2018-07-12
Delay after relese


olegrok:
> Example: queue.tube.list_of_sites:release(15, {delay=10})
> Note: in the above example, the delay option means "the task cannot be executed again for 10 seconds".

But.
```
q = queue.create_tube('tube', 'fifottl', { if_not_exists = true })
tarantool> q:put({'tablichka'}, { ttl = 600, ttr = 300 })
---
- [0, 'r', ['tablichka']]
...
tarantool> data = q:take()
tarantool> box.space.tube:select()
---
- - [0, 't', 1531394088891933, 600000000, 300000000, 0, 1531393773820542, ['tablichka']]
...
tarantool> q:release(0, {delay = 300})
---
- [0, '~', ['tablichka']]
...
tarantool> box.space.tube:select()
---
- - [0, '~', 1531394261199920, 600000300, 300000000, 0, 1531393773820542, ['tablichka']]
...
```

I suspect that `delay` now is in microseconds, but not seconds

Tarantool 1.9.0-61-gcd48321

PR #81, opened 2018-07-16
Fix ttl increase after delay

olegrok:
Close: https://github.com/tarantool/queue/issues/80

### DIFF
--- a/t/020-fifottl.t
+++ b/t/020-fifottl.t
@@ -2,7 +2,7 @@
 local fiber = require('fiber')
 
 local test = require('tap').test()
-test:plan(13)
+test:plan(14)
 
 local queue = require('queue')
 local state = require('queue.abstract.state')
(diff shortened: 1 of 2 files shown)


Totktonada:
@olegrok `touch` method need to be changed accordingly, yep?

Can you add a test case for both *ttl queue drivers (tests seems to be quite straighforward and that should not be a problem).

olegrok:
@Totktonada
No, `touch` works correctly. I sought correct behavior using `touch` before `release`.

I hope I can add tests a bit later

Totktonada:
`touch` has `increment` (`delta`) argument in seconds according to README and now it is just added to the `ttl` value of a task. `release` after your patch adds seconds * 10^6 to `ttl` value of a task. I think external API should use seconds and internal representation should use microseconds.

olegrok:
@Totktonada 
I added tests

Yes, internal representation of touch uses microseconds. And touch works correctly. See str. 116 in abstract.lua

Totktonada:
@olegrok Thanks! I got it, `touch` really works correctly. Now we have seconds for `delay` in `release` and microseconds for `delta` in `touch` in an API for drivers, but seconds everywhere in user-visible API. I think possible refactoring can go outside of scope on this PR.

LGTM. @olegrok, please proceed with merging it into master (squash + fast-forward / no squash + fast-forward / squash + merge commit / no squash + merge commit as you wish).

There are things I want to bring into focus as outcome:

* Drivers API is unchanged with cost of non-uniform drivers API (sec / microsec).
* External queue API was changed, but in the manner that the documentation already stated.
* It seems that the change does not need to involve data migration for old tasks (because we anyway cannot extract previous delays from ttl).

PR #86, opened 2019-01-24
implemented lifottl queue type

maxim-komar:
implemented lifottl queue type

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ align="right">
 * [Queue types](#queue-types)
   * [fifo \- a simple queue](#fifo---a-simple-queue)
   * [fifottl \- a simple priority queue with support for task time to live](#fifottl---a-simple-priority-queue-with-support-for-task-time-to-live)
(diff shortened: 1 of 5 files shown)


LeonidVas:
Hi. Thank you for the patch. Sorry for the big latency.
It's look like copy-paste of the fifottl driver with an overload of the take method (and this is expected). But WET is bad practice. I propose to pick out common code and write fifottl and lifottl as a wrapper around this (something like limfifottl relatively to fifottl).

LeonidVas:
Useful part of the PR is five lines of code, everything else - "copy-paste" from the fifottl driver. I agree that "copy-paste" methodology was used in the module for creating new drivers from existing ones. But support such code is difficult. I think DRY principle is a good practice and suggest to move common code to a separate file/module.
Back to the PR, I created an [issue](https://github.com/tarantool/queue/issues/135) to implement the lifo and lifottl drivers. Additionally, I want to note that the implementation of a new functionality always includes tests.
As for the unreasonably long feedback delays in the PR review: the party's policy has been changed and now part of the team's resource is allocated for this activity and it is expected such delays will gone. So, if you want, you can try to implement the task.

Issue #63, opened 2017-10-02
O(n^2) complexity in fifottl:take()


rtsisyk:
take() degrades to O(N^2) complexity where N is the number of expired tasks:

https://github.com/tarantool/queue/blob/3adda75efab42c8bf437cb554c81d269cfe1b27e/queue/abstract/driver/fifottl.lua#L265-L275

Affects Cloud.

kostja:
* Roman Tsisyk <notifications@github.com> [17/10/02 16:32]:
> take() degrades to O(N^2) complexity where N is the number of expired tasks:
> 
> https://github.com/tarantool/queue/blob/3adda75efab42c8bf437cb554c81d269cfe1b27e/queue/abstract/driver/fifottl.lua#L265-L275
> Affects Cloud.

I have received numerous complains about this, without exact trace to
the root cause. 

Let's fix asap.

-- 
Konstantin Osipov, Moscow, Russia, +7 903 626 22 32
http://tarantool.org - www.twitter.com/kostja_osipov

RunsFor:
I reproduced it under following conditions:

I put my laptop into sleep mode while it was about 50k tuples in fifottl tube and 10 fibers were constantly taking tasks from it from another instance over a net box connection.

I woke up laptop after ttl was already expired for all tasks and noticed that:

- Tarantool was draining 100% CPU
- Tarantool instance was not respond to any requests (http server, netbox connection, console)
- There was only mmap and munmap syscalls in strace (which means it was stuck inside one event loop)
- gdb was saying that tarantool constantly iterates over the space
- tarantool returned back to normal behaviour after about an hour (all expired tasks was deleted)

PR #88, opened 2019-03-26
fifottl: accelerate fifottl:take

olegrok:
* delete iteration over status index
* apply iteration with :pairs() over status index

### DIFF
--- a/queue/abstract/driver/fifottl.lua
+++ b/queue/abstract/driver/fifottl.lua
@@ -260,18 +260,16 @@ end
 
 -- take task
 function method.take(self)
-    local offset = 0
     local task = nil
-    while true do
-        task = self.space.index.status:select({state.READY},
-                {offset=offset, limit=1, iterator='GE'})[1]
-        if task == nil or task[i_status] ~= state.READY then
-            return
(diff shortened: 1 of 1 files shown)


Totktonada:
Changes I made (force-pushed):

* Rebased on top of current master.
* Added `Fixes #63`.

Totktonada:
Changes I made (force-pushed):

* Removed GE iterator and so status check inside the loop.

Totktonada:
@RunsFor Can you please check this fix on some more real cases? I mean that it does not break anything.

queue/abstract/driver/fifottl.lua
```diff
@@ -262,18 +262,17 @@ end
 function method.take(self)
     local offset = 0
```
rybakit: Seems like `offset` is not needed anymore?

Totktonada: Sure, thanks! Force-pushed.

olegrok:
@Totktonada 

I checked it for our project. I suppose that this changes don't break anything. Let's merge it.

Totktonada:
Thanks! Merged.

PR #89, opened 2019-03-26
travis-ci: update repositories and distros

Totktonada:


.travis.yml
```diff
@@ -15,16 +15,19 @@ env:
...
+      - OS=fedora DIST=28
+      - OS=fedora DIST=29
       - OS=ubuntu DIST=trusty
       - OS=ubuntu DIST=xenial
       - OS=ubuntu DIST=artful
       - OS=ubuntu DIST=bionic
-      - OS=debian DIST=wheezy
+      - OS=ubuntu DIST=cosmic
       - OS=debian DIST=jessie
       - OS=debian DIST=stretch
```
rybakit:
```suggestion
      - OS=debian DIST=buster
      - OS=debian DIST=sid

```

Totktonada: At least we need tarantool on sid in our repository (and appropriate testing pipelines) to use it as a dependency in a module. So I would expect until https://github.com/tarantool/tarantool/issues/4037 will be resolved or declined and only then going to update modules.

rybakit: I see, let's wait then :)

PR #90, opened 2019-03-26
Fix 090 test on MAC

olegrok:
Before it failed with error:
2019-03-26 22:52:57.418 [97087] main/101/090-grant-check.t F> ./t/090-grant-check.t:64: assign to undeclared variable 'queue'

Now this test is passed

### DIFF
--- a/t/090-grant-check.t
+++ b/t/090-grant-check.t
@@ -58,6 +58,7 @@ test:test('check for space grants', function(test)
     tube:drop()
 end)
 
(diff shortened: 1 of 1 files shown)


Totktonada:
Rewrote with `rawset()` (to keep all test case related actions within its code block), rewrote the commit message (it is about the strict mode, not about Mac OS) and pushed to master as 19746d38f6246d269c47024d318b53142d474646.

PR #91, opened 2019-07-11
Add limited fifo ttl queue

AndreiNepsha:


queue/abstract/driver/limfifottl.lua
```diff
@@ -0,0 +1,42 @@
...
+        while true do
+            local tube_size = self.space:count()
+            if tube_size < state.capacity or state.capacity == 0 then
+                return state.parent.put(self, data, opts)
+            else
+                if timeout == 0 then
+                    return nil
+                end
+
+                local started = fiber.time64()
```
Kasen: I think time64 is redundant, and I prefer to use fiber.time().

queue/abstract/driver/limfifottl.lua
```diff
@@ -0,0 +1,42 @@
...
+        parent = fifottl.new(space, on_task_change, opts)
+    }
+
+    -- put task in space
+    local put = function (self, data, opts)
+        local timeout = opts.timeout or 0
+        timeout = timeout * 1000000
+
+        while true do
+            local tube_size = self.space:count()
```
Kasen: self.space:len() is more preferable here, because it is faster, but has less precision.

queue/abstract/driver/limfifottl.lua
```diff
@@ -0,0 +1,42 @@
...
+function tube.new(space, on_task_change, opts)
+    local state = {
+        capacity = opts.capacity or 0,
+        parent = fifottl.new(space, on_task_change, opts)
+    }
+
+    -- put task in space
+    local put = function (self, data, opts)
+        local timeout = opts.timeout or 0
+        timeout = timeout * 1000000
```
Kasen: If fiber.time64() will be changed on fiber.time(), then this multiplication will be not needed.

queue/abstract/driver/limfifottl.lua
```diff
@@ -0,0 +1,42 @@
...
+            local tube_size = self.space:count()
+            if tube_size < state.capacity or state.capacity == 0 then
+                return state.parent.put(self, data, opts)
+            else
+                if timeout == 0 then
+                    return nil
+                end
+
+                local started = fiber.time64()
+                fiber.yield()
```
Kasen: fiber.yield() may cause a big cpu load, so I prefer to use fiber.sleep(0.01)

Issue #92, opened 2019-07-25
Full scan in `queue.statistics()`


derElektrobesen:
```lua
  1 local function build_stats(space)
  2     local stats = {tasks = {}, calls = {
  3         ack  = 0, bury  = 0, delete  = 0,
  4         kick = 0, put   = 0, release = 0,
  5         take = 0, touch = 0,
  6         -- for *ttl queues only
  7         ttl  = 0, ttr   = 0, delay   = 0,
  8     }}
  9
 10     local st = rawget(queue.stat, space) or {}
 11     local idx_tube = 1
 12
 13     -- add api calls stats
 14     for name, value in pairs(st) do
 15         if type(value) ~= 'function' and name ~= 'done' then
 16             stats['calls'][name] = value
 17         end
 18     end
 19
 20     -- add total tasks count
 21     stats['tasks']['total'] = box.space[space].index[idx_tube]:count()
 22
 23     -- add tasks by state count
 24     for i, s in pairs(state) do
 25         stats['tasks'][i:lower()] = box.space[space].index[idx_tube]:count(s)
 26     end
 27     stats['tasks']['done'] = st.done or 0
 28
 29     return stats
 30 end
```

This function calls `count()` 6 times on each `queue.statistics()`. This could be a reason of very aggressive CPU usage if there are very many tasks in the queue.

olegrok:
In de0a346f17e690aed19384577ed5157476f8fa3a we removed redundant fullscan from build stats

We can think about implementing a view that build stats on start and increment it then (see https://github.com/tarantool/queue/pull/93)

PR #93, opened 2019-07-25
Remove redundant full-scan on stat build

olegrok:
Need for #92

### DIFF
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -536,13 +536,16 @@ local function build_stats(space)
         end
     end
 
-    -- add total tasks count
-    stats['tasks']['total'] = box.space[space].index[idx_tube]:count()
-
+    local total = 0
     -- add tasks by state count
     for i, s in pairs(state) do
-        stats['tasks'][i:lower()] = box.space[space].index[idx_tube]:count(s)
+        local st = i:lower()
+        stats['tasks'][st] = box.space[space].index[idx_tube]:count(s)
+        total = total + stats['tasks'][st]
     end
+
+    -- add total tasks count
+    stats['tasks']['total'] = total
     stats['tasks']['done'] = st.done or 0
 
     return stats


olegrok:
> Hi! Thanks for the patch!
> Look good, but I am proposing to add some benchmarks with results.

The simplest variant. I use vynil as engine, because it allows to get more indicative results
```lua
local clock = require('clock')
local json = require('json')
local queue = require('queue')

box.cfg {
    vinyl_memory = 100000,
    log_level = 3,
    vinyl_write_threads = 5,
}

local tube = queue.create_tube('bench', 'fifo', {engine = 'vinyl'})

local count = 1e5


local task = string.rep('x', 10)
for _ = 1, count do
    tube:put({task})
end

for _ = 1, count / 2 do
    tube:take()
end

print('Inserted:', count)

collectgarbage()
collectgarbage()

local start = clock.time()
local stats = queue.stats()
local result = clock.time() - start
print(result, json.encode(stats))

box.snapshot()
os.exit(0)
```

Before: 2.53s
After: 2.13s

LeonidVas:
Hello. Thanks for the benchmark. It's look like the patch improves performance(~16%). And LGTM. But I think that calculating statistics during a work of the process  looks better. This will allow to have relevant data at hand.

olegrok:
> Hello. Thanks for the benchmark. It's look like the patch improves performance(~16%). And LGTM. But I think that calculating statistics during a work of the process looks better. This will allow to have relevant data at hand.

Yes, I agree, I think we could call current `stats` implementation once at start then increment counter for every operations. But it should be done in separate patch.

As it was done here: https://github.com/moonlibs/xqueue/blob/master/xqueue.lua#L1326

Totktonada:
I think the fix is okay. We can left the issue #92 open and implement a kind of materizlized view later.

Totktonada:
Oleg, please, don't do this anymore. I had the plan to release all critical fixes in one release (it'll as stable as possible), but perf. improvements in another one.

Totktonada:
Restricted `master` branch.

olegrok:
Ok, sorry. I didn't know about your plan and understand your "approve" as "ready to push"

PR #97, opened 2019-09-06
utubettl: remove unnecessary on_task_change call

iskandarov-egor:


queue/abstract/driver/utubettl.lua
```diff
@@ -290,24 +290,24 @@ function method.take(self)
...
 
-local function process_neighbour(self, task, operation)
-    self:on_task_change(task, operation)
-    if task ~= nil then
-        local neighbour = self.space.index.utube:min{state.READY, task[i_utube]}
-        if neighbour ~= nil and neighbour[i_status] == state.READY then
-            self:on_task_change(neighbour)
-        end
+local function wake_neighbour(self, utube)
+    local neighbour = self.space.index.utube:select({state.READY, utube}, {limit=1})[1]
```
LeonidVas: I think using min() with checks is more correct (with link to https://github.com/tarantool/tarantool/issues/3167)

queue/abstract/driver/utubettl.lua
```diff
@@ -290,24 +290,24 @@ function method.take(self)
...
     local task = self.space:get(id)
-    self.space:delete(id)
     if task ~= nil then
+        self.space:delete(id)
+        if task[i_status] == state.TAKEN then
+            wake_neighbour(self, task[i_utube])
+        end
         task = task:transform(i_status, 1, state.DONE)
-        return process_neighbour(self, task, 'delete')
+        self:on_task_change(task, 'delete')
```
LeonidVas: The self:on_task_change call should be before wake_neighbour (old logic).

queue/abstract/driver/utubettl.lua
```diff
@@ -324,9 +324,9 @@ function method.release(self, id, opts)
             { '=', i_next_event, event_time(opts.delay) },
             { '+', i_ttl, time(opts.delay) }
         })
-        if task ~= nil then
-            return process_neighbour(self, task, 'release')
-        end
+        self:on_task_change(task, 'release')
```
LeonidVas: Check if task ~= nil then

queue/abstract/driver/utubettl.lua
```diff
@@ -341,9 +341,10 @@ end
 function method.bury(self, id)
     local task = self.space:update(id, {{ '=', i_status, state.BURIED }})
     if task ~= nil then
-        return process_neighbour(
-            self, task:transform(i_status, 1, state.BURIED), 'bury'
-        )
+        wake_neighbour(self, task[i_utube])
+        task = task:transform(i_status, 1, state.BURIED)
```
LeonidVas: The self:on_task_change call should be before wake_neighbour (old logic).

queue/abstract/driver/utubettl.lua
iskandarov-egor: aren't they equivalent?

iskandarov-egor:
> It seems it is possible that a next task of utube may be taken before the previous one is acked. It violates guarantees of strict odrdering within a utube.

How? Does my patch introduce this bug, or does it already exist?

queue/abstract/driver/utubettl.lua
LeonidVas: I meant that the behavior of max() was different from expected and I see no reason to replace min() to select(). I agree with you that the default iterator is ALL(same as GE) and the results will be equivalent. But generally min () is used in utubettl driver and IMHO: it better reflects the idea. I may be wrong. What is the reason for choosing select()?

iskandarov-egor: from the 1.10 docs:
> In a future version of Tarantool, index:min(key value) will return nothing if key value is not equal to a value in the index.

so we cannot be sure that `neighbour` will be from the same utube on tarantool 1.10

LeonidVas: Hmm, as I understand you say about https://github.com/tarantool/tarantool/commit/8bbd4a6803a77673c7f146b6946dfcb9c814dce4 . Looks like you're right! Please add a comment. If you add a test, it will be great!

LeonidVas: > I meant that the behavior of max() was different from expected and I see no reason to replace min() to select(). I agree with you that the default iterator is ALL(same as GE) and the results will be equivalent.
The default iterator is 'EQ'.

LeonidVas:
I had a conversation with @Totktonada and your code is generally ok. Now it needs to be decorated in accordance with the general points https://github.com/tarantool/tarantool/wiki/Code-review-procedure (commit message, comments, tests if possible). You can do this work yourself or forward it to us. Which pill do you choose?

artur-barsegyan:
> I had a conversation with @Totktonada and your code is generally ok. Now it needs to be decorated in accordance with the general points https://github.com/tarantool/tarantool/wiki/Code-review-procedure (commit message, comments, tests if possible). You can do this work yourself or forward it to us. Which pill do you choose?

I think we should make an optimistic merge in the future :)

Totktonada:
> I think we should make an optimistic merge in the future :)

Only sceptics can keep quality on the production level. Sometimes you're lucky and 'rapid development' approach seems to work, but it does not scale to a time period like several years. There are ways to exploit this rule: say, hire a large QA team and perform thorough pre-release testing. But it is costly.

It is based on my experience, which is not a gold standard. Maybe your experience in long living projects is different.

LeonidVas:
Has been updated and pushed to master.

PR #99, opened 2019-09-18
fix fifottl_fiber_iteration error handling

### LINKED ISSUE #96: Infinite loop on (possible) ttr return of a buried task to the dropped queue
Tarantool Enterprise 1.10.3-89-g412e943, queue 1.0.3
 
Bug can be reproduced by the following script

```
local queue = require('queue')
local fiber = require('fiber')

box.cfg{}
queue.create_tube('task_line', 'fifottl', {ttr = 10, if_not_exist = true})
queue.tube.task_line:put({task = "task"})
queue.tube.task_line:put({task1 = "task"})
local taken_task = queue.tube.task_line:take(0)

local tube_id = box.space._queue:select{tube_name = 'task_line'}[1].tube_id
local tasks_taken = box.space._queue_taken:select{tube_id = tube_id}

for _, task in ipairs(tasks_taken) do
    queue.tube.task_line:bury(task.task_id)
end
queue.tube.task_line:drop()
fiber.sleep(20)
```
image

After ~`ttr` seconds of sleep tarantool instance start to use ~100% of CPU, became unavailable for connection by net.box and can't be stopped by Ctrl-C.

PersDep:
fifottl_fiber pcalls fifottl_fiber_iteration and then handles
result. Returned error code comparison with box.error.READONLY
was not done properly. Now it is fixed.

Closes #96

### DIFF
--- a/queue/abstract/driver/fifottl.lua
+++ b/queue/abstract/driver/fifottl.lua
@@ -164,7 +164,7 @@ local function fifottl_fiber(self)
     while true do
         if not box.cfg.read_only then
             local stat, err = pcall(fifottl_fiber_iteration, self, processed)
-            if not stat and not err.code == box.error.READONLY then
+            if not stat and not (err.code == box.error.READONLY) then
                 log.error("error catched: %s", tostring(err))
                 log.error("exiting fiber '%s'", fiber.name())
                 return 1
--- a/t/020-fifottl.t
+++ b/t/020-fifottl.t
@@ -2,7 +2,7 @@
 local fiber = require('fiber')
 
 local test = require('tap').test()
-test:plan(14)
+test:plan(15)
 
 local queue = require('queue')
 local state = require('queue.abstract.state')
@@ -243,6 +243,23 @@ test:test('ttl after delay test', function(test)
     local task = box.space.test_ttl_release:get(0)
     test:is(task.ttl, (TTL + DELTA) * 1000000, 'check TTL after release')
     test:is(task.ttr, TTR * 1000000, 'check TTR after release')
+end)
+
+-- gh-96: infinite loop after dropping a tube with a burried task
+test:test('buried task in a dropped queue', function(test)
+    test:plan(1)
+
+    local TASK_ID = 1
+    local tube = queue.create_tube('test_drop_with_burried', 'fifottl',
+        {ttr = 0.1, if_not_exist = true})
+
+    tube:put({foo = 'bar'})
+    local task = tube:take(0)
+    tube:bury(task[TASK_ID])
+
+    tube:drop()
+    fiber.sleep(0.2)
+    test:ok(true, 'queue does not hang')
 end)
 
 


queue/abstract/driver/fifottl.lua
```diff
@@ -164,7 +164,7 @@ local function fifottl_fiber(self)
     while true do
         if not box.cfg.read_only then
             local stat, err = pcall(fifottl_fiber_iteration, self, processed)
-            if not stat and not err.code == box.error.READONLY then
+            if not stat and not (err.code == box.error.READONLY) then
```
olegrok: Thanks, good catch.

But could you add a relevant test case?

PersDep: 👌

t/020-fifottl.t
```diff
@@ -245,6 +245,22 @@ test:test('ttl after delay test', function(test)
     test:is(task.ttr, TTR * 1000000, 'check TTR after release')
 end)
 
+test:test('buried task in dropped queue', function(test)
+    test:plan(1)
+    queue.create_tube('task_line', 'fifottl', {ttr = 1, if_not_exist = true})
```
Totktonada: I would use 0.1 and 0.2 seconds. I think that is enough.

t/020-fifottl.t
```diff
@@ -245,6 +245,22 @@ test:test('ttl after delay test', function(test)
     test:is(task.ttr, TTR * 1000000, 'check TTR after release')
 end)
 
+test:test('buried task in dropped queue', function(test)
```
Totktonada: I would write it in simpler way:

```lua
-- gh-96: infinite loop after drop a tube with burried task
test:test('buried task in dropped queue', function(test)
    test:plan(1)

    local TASK_ID = 1
    local tube = queue.create_tube('test_drop_with_burried', 'fifottl',
        {ttr = 0.1, if_not_exist = true})

    tube:put({foo = 'bar'})
    local task = tube:take(0)
    tube:bury(task[TASK_ID])

    tube:drop()
    fiber.sleep(0.2)
    test:ok(true, 'queue does not hang')
end)
```

t/020-fifottl.t
```diff
@@ -245,6 +245,22 @@ test:test('ttl after delay test', function(test)
...
+    queue.tube.task_line:put({task1 = "task"})
+    local taken_task = queue.tube.task_line:take(0)
+    local tube_id = box.space._queue:select{tube_name = 'task_line'}[1].tube_id
+    local tasks_taken = box.space._queue_taken:select{tube_id = tube_id}
+    for _, task in ipairs(tasks_taken) do
+        queue.tube.task_line:bury(task.task_id)
+    end
+    queue.tube.task_line:drop()
+    fiber.sleep(2)
+    test:is(table.getn(queue.tube), 0, 'check that queue.tube is empty')
```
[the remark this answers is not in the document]
PersDep: Yes, I actually used that check wrong way.
Right, it should be simplified to simple hang check

t/020-fifottl.t
PersDep: I agree, will fix

t/020-fifottl.t
PersDep: This really is much simpler. We don't need 2 task to reproduce, that was a mistake.

PR #102, opened 2019-11-29
Publish rockspec automatically

printercu:


.travis.yml
```diff
@@ -8,106 +8,79 @@ cache:
...
-  - ls -l build/
+_deploy_to_packagecloud: &deploy_to_packagecloud
+  provider: packagecloud
+  username: tarantool
+  token: $PACKAGECLOUD_TOKEN
+  dist: $OS/$DIST
+  package_glob: build/*.{rpm,deb}
+  skip_cleanup: true
+
+# Deploy packages to PackageCloud from master branch
```
Totktonada: Nit: ` from master branch` is redundant here.

printercu:
I've split deployment from tests because deployment includes building packages with packpack. This takes quite long. Now it runs tests for 3 Tarantool versions in `push` travis job, if it succeed it runs tests in `pr` travis job and only then starts building packages.

I've also added commit for publishing rockspecs.

Totktonada:
I wrote it in assumption that we run tests in packpack jobs itself, but now I see that it is so only in Debian / Ubuntu jobs, while test are not run in CentOS / Fedora jobs. Okay for splitting stages so.

Issue #85, opened 2018-11-21
Cannot :ack task after implicit net.box reconnect


RunsFor:
**Prerequisite**

1. Server which stores tasks using tarantool/queue with custom driver based on utubettl
2. Client which takes tasks from the server through `net.box` connection

**Scenario**

1. Client takes a task through `net.box` connection and start processing it
2. Network glitch (or another issue) occur and `net.box` decides to reconnect
3. Client finishes processing task and decides to `ack` the task.
4. Server responds to `ack` command with error - `Task was not taken in the session`
5. After `ttr` delay task returned to the `READY` state

**Problem description**

This "`ack` error" happen because tarantool/queue [locks](https://github.com/tarantool/queue/blob/master/queue/abstract.lua#L313) "taken" task to `box.session.id()`. That means that not only the same client must `take` and `ack` the same task, but it has to be done inside the same `net.box` connection, since `box.session.id()` updates after implicit reconnect. That implies that client has no way to handle such error from the server and retry `ack`.

**Possible solutions**

There is a possible workaround to eliminate this kind of errors is to implement a simple buffer (`fiber.channel`) on the server with client `ack` commands and let another fiber to call an actual `queue:ack`. But this is not very handy, if you want to handle `ack` response on the client.

Another way to approach this is to use different `id` to lock a task. This `id` should determine the same client even through implicit `net.box` reconnect calls. This id may be set explicitly through some registration process (api breaking change) or implicitly using some tarantool client information (if exists).

Is this reasoning correct? Are there any different approaches/workarounds?

Totktonada:
A note from @printercu: if we'll allow reconnects, then we possibly should not automatically release tasks in on_disconnect trigger.

printercu:
Maybe this should be an option because it depends on queue type: I use `ttr` to rerun stale tasks and ping queue continuously from active worker.

The other related bug is that task can not be handled from different fibers within single process: `box.session.id` returns different values from different fibers.

PR #106, opened 2020-01-27
Release all taken tasks on start

### LINKED ISSUE #66: Taken task is not released back to 'ready' after Tarantool reboot
How to reproduce:

1) Get into the docker container and add a task:
```
docker-compose exec tnt1 tarantoolctl connect /var/run/tarantool/tarantool.sock
queue.tube.default:put({test=true})
---
- [0, 'r', {'test': true}]
...
```

2) Take a task:

```
queue.tube.default:take()
---
- [0, 't', {'test': true}]
...
```

3) Restart container:

```
docker-compose restart tnt1
```

4) Check task status:

```
queue.tube.default:peek(0)
---
- [0, 't', {'test': true}]
...
```

The issue can be reproduced with connectors:

https://github.com/igorcoding/asynctnt-queue:

```python
import asyncio
import asynctnt
import asynctnt_queue
import sys

async def run():
    conn = asynctnt.Connection(host='tnt1', port=3301, username='tnt1', password='tnt1')
    await conn.connect()

    queue = asynctnt_queue.Queue(conn)
    test_tube = queue.tube('default')

    # Retrieve a task from queue
    task = await test_tube.take(1)

    print(task)

    # Restart Tarantool here
    sys.stdin.readline()

    await conn.disconnect()

loop = asyncio.get_event_loop()
loop.run_until_complete(run())
```

https://github.com/tarantool-php/queue:

```php
<?php

$conn = new \Tarantool\Client\Connection\StreamConnection('tcp://127.0.0.1:3301');
$client = new \Tarantool\Client\Client($conn, new \Tarantool\Client\Packer\PurePacker());
$client->authenticate('tnt1', 'tnt1');

$queue = new \Tarantool\Queue\Queue($client, 'default');
$task = $queue->take(1);

var_dump($task);

readline('Restart Tarantool and press a key to continue');
```

Here is the simplified instance configuration:

```lua

box.cfg {
    listen = 3301,
    log_level = 5
}

local config = {
    user = 'tnt1',
    password = 'tnt1'
}

if not box.schema.user.exists(config.user) then
    box.schema.user.create(config.user, {password = config.password})
    box.schema.user.grant(config.user, 'read,write,execute', 'universe', nil)
end

queue = require('queue')
box.once('foobar:v0.1.0', function()
    queue.create_tube('default', 'fifottl', {if_not_exists = true})
end)
```

LeonidVas:
If some tasks have been taken and don't released before shutdown
of the tarantool instance (for example: tarantool instance has been killed)
such task go to 'hung' state (noone can take the task now).
So, we must release all taken tasks on start of the queue module.

Fixes #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,6 +637,7 @@ If there are no 'ready' tasks in the queue, returns nil.
 * `tube:bury(task_id)` - buries a task
 * `tube:kick(count)` - digs out `count` tasks
 * `tube:peek(task_id)` - return the task state by ID
+* `tube:tasks_by_state(task_state)` - return the iterator to tasks in a certain state
 * `tube:truncate()` - delete all tasks from the tube. Note that `tube:truncate`
 must be called only by the user who created this tube (has space ownership) OR
 under a `setuid` function. Read more about `setuid` functions
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -63,6 +63,26 @@ queue.driver = {
     limfifottl  = require('queue.abstract.driver.limfifottl')
 }
 
+local function tube_release_all_tasks(tube)
+    local prefix = ('queue: [tube "%s"] '):format(tube.name)
+
+    -- We lean on stable iterators in this function.
+    -- https://github.com/tarantool/tarantool/issues/1796
+    if not qc.check_version({1, 7, 5}) then
+        log.error(prefix .. 'no stable iterator support: skip task releasing')
+        log.error(prefix .. 'some tasks may stuck in taken state perpetually')
+        log.error(prefix .. 'update tarantool to >= 1.7.5 or take the risk')
+    end
+
+    log.info(prefix .. 'releasing all taken task (may take a while)')
+    local released = 0
+    for _, task in tube.raw:tasks_by_state(state.TAKEN) do
+        tube.raw:release(task[1])
+        released = released + 1
+    end
+    log.info(prefix .. ('released %d tasks'):format(released))
+end
+
 -- tube methods
 local tube = {}
 
@@ -515,7 +535,11 @@ function method.start()
         })
     end
 
-    _queue:pairs():each(recreate_tube)
+    for _, tube_tuple in _queue:pairs() do
+        local tube = recreate_tube(tube_tuple)
+        -- gh-66: release all taken tasks on start
+        tube_release_all_tasks(tube)
+    end
 
     session.on_disconnect(queue._on_consumer_disconnect)
     return queue
--- a/queue/abstract/driver/fifo.lua
+++ b/queue/abstract/driver/fifo.lua
@@ -121,6 +121,11 @@ function method.peek(self, id)
     return self.space:get{id}
 end
 
+-- get iterator to tasks in a certain state
+function method.tasks_by_state(self, task_state)
+    return self.space.index.status:pairs(task_state)
+end
+
 function method.truncate(self)
     self.space:truncate()
 end
--- a/queue/abstract/driver/fifottl.lua
+++ b/queue/abstract/driver/fifottl.lua
@@ -348,6 +348,11 @@ function method.peek(self, id)
     return self.space:get{id}
 end
 
+-- get iterator to tasks in a certain state
+function method.tasks_by_state(self, task_state)
+    return self.space.index.status:pairs(task_state)
+end
+
 function method.truncate(self)
     self.space:truncate()
 end
--- a/queue/abstract/driver/utube.lua
+++ b/queue/abstract/driver/utube.lua
@@ -151,6 +151,11 @@ function method.peek(self, id)
     return self.space:get{id}
 end
 
+-- get iterator to tasks in a certain state
+function method.tasks_by_state(self, task_state)
+    return self.space.index.status:pairs(task_state)
+end
+
 function method.truncate(self)
     self.space:truncate()
 end
--- a/queue/abstract/driver/utubettl.lua
+++ b/queue/abstract/driver/utubettl.lua
@@ -370,6 +370,11 @@ function method.peek(self, id)
     return self.space:get{id}
 end
 
+-- get iterator to tasks in a certain state
+function method.tasks_by_state(self, task_state)
+    return self.space.index.status:pairs(task_state)
+end
+
 function method.truncate(self)
     self.space:truncate()
 end
--- a/t/000-init.t
+++ b/t/000-init.t
@@ -43,5 +43,5 @@ test:test('access to queue after box.cfg{}', function(test)
 end)
 
 tnt.finish()
-os.exit(test:check() == true and 0 or -1)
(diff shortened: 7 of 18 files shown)


LeonidVas:
Test fails are not related with changes

queue/abstract.lua
```diff
@@ -511,7 +511,20 @@ function method.start()
...
+        local space_name = tube_tuple[3]
+        tube = recreate_tube(tube_tuple)
+        local space = box.space[space_name]
+        for _, task_tuple in space:pairs() do
+            -- Release all taken tasks on start
+            -- See https://github.com/tarantool/queue/issues/66
+            -- for more information
+            local tid, task_state = task_tuple[1], task_tuple[2]
+            if task_state == state.TAKEN then
+                tube.raw:release(tid)
```
Totktonada: Sooner or later we'll return to #85 and I guess releasing tasks after restart will need to be optional after that. But not now, just noted.

queue/abstract.lua
```diff
@@ -511,7 +511,20 @@ function method.start()
         })
     end
 
-    _queue:pairs():each(recreate_tube)
+    for _, tube_tuple in _queue:pairs() do
+        local space_name = tube_tuple[3]
+        tube = recreate_tube(tube_tuple)
+        local space = box.space[space_name]
+        for _, task_tuple in space:pairs() do
```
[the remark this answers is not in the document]
LeonidVas: You are so right. I'll redo the patch. Look like build_stats must be updated too.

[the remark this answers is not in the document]
LeonidVas: Complete

Totktonada:
> Replace exit code in case of error from 255 to 1

Let's add 'test' prefix.

queue/abstract.lua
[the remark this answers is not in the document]
Totktonada: > I would check a version and give a warning if we run on tarantool version w/o stable iterators: [tarantool/tarantool#1796](https://github.com/tarantool/tarantool/issues/1796) . If so, just don't release tasks and give the warning.

Added in 'FIXUP' commit.

Totktonada:
I put two 'FIXUP' commits.

The first one squashed duplicated testing code. It also adds the test for limfifottl.

The second one adds more logs and checking of tarantool version (as was proposed before). It also fixes the potential problem: `tube = recreate_tube(tube_tuple)` rewrotes upvalue tube (moduel level variable) — added `local`. Removed forgotten `space_name` variable. Moved the releasing logic into a function (the reason is mostly to skip it easily by a condition, in light of #85).

After the fixups I'm happy with the change. If you're too, please, squash the fixups into the base commit and I'll proceed with the patchset.

queue/abstract.lua
```diff
@@ -63,6 +63,26 @@ queue.driver = {
     limfifottl  = require('queue.abstract.driver.limfifottl')
 }
 
+local function tube_release_all_tasks(tube)
+    local prefix = ('queue: [tube "%s"] '):format(tube.name)
+
+    -- We lean on stable iterators in this function.
+    -- https://github.com/tarantool/tarantool/issues/1796
+    if not qc.check_version({1, 7, 5}) then
+        log.error(prefix .. 'no stable iterator support: skip task releasing')
```
LeonidVas: It's look like warning or info, not error

queue/abstract.lua
```diff
@@ -63,6 +63,26 @@ queue.driver = {
...
+        log.error(prefix .. 'some tasks may stuck in taken state perpetually')
+        log.error(prefix .. 'update tarantool to >= 1.7.5 or take the risk')
+    end
+
+    log.info(prefix .. 'releasing all taken task (may take a while)')
```
LeonidVas: In my opinion,  it's quick action and log is useless.

LeonidVas:
About common tests. I think about it, but I'm disagree with proposed concept. If look at drivers tests it have many copy-paste code. My point of view: "We must to have some common tests which will be run for all/some drivers from list".  And this is not related to the patch.
limfifottl don't need the test, because it is a wrap around fifottl.

queue/abstract.lua
Totktonada: I'll ask an order for some hot instance from our operation team.

queue/abstract.lua
Totktonada: queue is not operational in the case. It is the error in my understanding.

Totktonada:
> We must to have some common tests which will be run for all/some drivers from list

The test case should be either referenced from a certail test file (like `020-fifottl.t`) or it should be obvious from names that certain part of cases is in a separate file (say, `ddd-driver_api.t`).

We discussed it more with Leonid and he want to keep current way for now and make deduplication separately. I would do it like proposed above, but this way is okay too.

> limfifottl don't need the test, because it is a wrap around fifottl.

It is implementation detail, but we test public API of a driver. So I would left the test here.

queue/abstract.lua
Totktonada: I got the answer that they have small amount of taken tasks (on utubettl driver).

However I suspect that different cases are possible. Don't sure here, to be honest.

Feel free to remove the print.

LeonidVas:
Updated

LeonidVas:
limfifottl does not duplicate fifottl tests. So, let's continue to go that way

README.md
```diff
@@ -637,6 +637,7 @@ If there are no 'ready' tasks in the queue, returns nil.
 * `tube:bury(task_id)` - buries a task
 * `tube:kick(count)` - digs out `count` tasks
 * `tube:peek(task_id)` - return the task state by ID
+* `tube:tasks_by_state(task_state)` - return the iterator to tasks in a certain state
```
rybakit: Should [tube.grant](https://github.com/tarantool/queue/blob/de0a346f17e690aed19384577ed5157476f8fa3a/queue/abstract.lua#L267-L277) be updated to grant proper persissions to this new function?

rybakit: Nit: *Almost* all other public function names begin with a verb (except `stats`). Maybe we should stick to that convention and rename this function to `get_tasks_by_state()` or `get_by_state()`? Wdyt?

LeonidVas: No. It's about driver API, not about tube API.

LeonidVas: In this section of README read 'driver', don't read tube.

Totktonada: > Should [tube.grant](https://github.com/tarantool/queue/blob/de0a346f17e690aed19384577ed5157476f8fa3a/queue/abstract.lua#L267-L277) be updated to grant proper persissions to this new function?

We call it only on start and don't provide a way to call it (from public API point of view), so I guess no.

Totktonada: I would even name it just `:pairs(<...>)` with ability to provide different arguments. Now it'll be `{status = <...>}`.

Totktonada: Let's keep the current method to fix the bug. If we'll agree how to better name it in a future — let's change then.

PR #107, opened 2020-01-30
Fix task release in session disconnect

### LINKED ISSUE #103: bug: wrong session id in on_disconnect() trigger
Tarantool had a bug with on_disconnect triggers, see https://github.com/tarantool/tarantool/issues/4627

Affected tarantool versions:

* All 1.10* before 1.10.4-52-gcf99b51cf (exclusive).
* All 2.1.
* All 2.2* before 2.2.1-124-g581f3674e (exclusive).
* All 2.3* before 2.3.0-236-g6da9d3953 (exclusive).

The module uses on_disconnect() triggers and may call box.session.id() after implicit yield (a `:delete()` call).

We possibly should work around this problem in the module itself, despite that it was fixed in recent tarantool versions.

The fix would be obvious: save a session_id at beginning of `_on_consumer_disconnect()` and pass it to `:release()` at the end (or, better, to an internal release function with this extra option to protect a user from using it). However I stuck for a while trying to write a test case that will show a bad behaviour of queue just using its public API.

Reported by @Mons.

LeonidVas:
Fixes #103

### DIFF
--- a/t/110-disconnect-trigger-check.t
+++ b/t/110-disconnect-trigger-check.t
@@ -1,0 +1,98 @@
+#!/usr/bin/env tarantool
+
+local fiber = require('fiber')
+local netbox = require('net.box')
+local os = require('os')
+local queue = require('queue')
+local tap = require('tap')
+local tnt = require('t.tnt')
+
+local tube
+local test = tap.test('lost a session id after yield')
+
+-- The test cases are in check_result().
+test:plan(2)
+
+-- Verify that _queue_taken space is empty.
+local function check_result()
+    if tube == nil then
+        os.exit(1)
+    end
+
+    -- tube:drop() is most simple way to check that _queue_taken
+    -- is empty. It give an error if it is not so.
+    local ok, res = pcall(tube.drop, tube)
+    test:is(ok, true, 'drop empty queue')
+    test:is(res, true, 'tube:drop() result is true')
+
+    tnt.finish()
+    os.exit(test:check() and 0 or 1)
+end
+
+-- Yield in queue's on_disconnect trigger (which handles a client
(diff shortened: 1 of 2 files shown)


t/110-my_test.t
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env tarantool
```
olegrok: Filename is `t/110-my_test.t`

Please, consider changing it to more appropriate

LeonidVas: D'oh. Thanks.

LeonidVas: Fixed

t/110-disconnect-trigger-check.t
```diff
@@ -0,0 +1,66 @@
...
+    if tube == nil then
+        os.exit(-1)
+    end
+
+    local ok, res = pcall(tube.drop, tube)
+    test:is(ok, true, 'drop empty queue')
+    test:is(res, true, 'tube:drop() result is true')
+
+    tnt.finish()
+    os.exit(test:check() == true and 0 or -1)
```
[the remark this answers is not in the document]
Totktonada: Resolved as not important.

Totktonada:
Ubuntu Cosmic fails: the fail is expected, because the distro reached EOL:

```
sudo apt-get update > /dev/null
E: The repository 'http://security.ubuntu.com/ubuntu cosmic-security Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu cosmic Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu cosmic-updates Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu cosmic-backports Release' does not have a Release file.
```

PR #108, opened 2020-01-30
Add a check of the session state before take a task

### LINKED ISSUE #104: bug: second take() hangs after disconnect during take()
@Mons describes the following case when take() may hang.

Take a task from a connector (say, net.box), disconnect and connect again, try to take this task again. It is stated that the second take will (may?) hang.

It seems that something like this was fixed in xqueue here: https://github.com/moonlibs/xqueue/commit/eb501508c9ffc7ca5f56613a41458d2bb3c0843a

Let's verify it, fix and cover with a test.

LeonidVas:
We need to check of a session state before take a task after wait()
because session maybe in disconnecting state and task will be
hang in a take state after the session will be disconnected

Fixes #104

### DIFF
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -73,6 +73,7 @@ function tube.put(self, data, opts)
 end
 
 local conds = {}
+local releasing_sessions = {}
 
 function tube.take(self, timeout)
     timeout = time(timeout or TIMEOUT_INFINITY)
@@ -93,6 +94,13 @@ function tube.take(self, timeout)
         conds[fid]:wait(tonumber(timeout) / 1000000)
         conds[fid]:free()
         box.space._queue_consumers:delete{ sid, fid }
+
+        -- We don't take a task if the session is in a
+        -- disconnecting state.
+        if releasing_sessions[fid] then
+            releasing_sessions[fid] = nil
+            return nil
+        end
 
         task = self.raw:take()
 
@@ -356,6 +364,7 @@ function method._on_consumer_disconnect()
         box.space._queue_consumers:delete{ waiter[1], waiter[2] }
         local cond = conds[waiter[2]]
         if cond then
+            releasing_sessions[waiter[2]] = true
             cond:signal(waiter[2])
         end
     end
--- a/t/120-take-task-after-reconnect.t
+++ b/t/120-take-task-after-reconnect.t
@@ -1,0 +1,62 @@
+#!/usr/bin/env tarantool
+
+local fiber = require('fiber')
+local netbox = require('net.box')
+local os = require('os')
+local queue = require('queue')
+local tap = require('tap')
+local tnt = require('t.tnt')
+
+
+local test = tap.test('take a task after reconnect')
+test:plan(1)
+
+local listen = 'localhost:1918'
+tnt.cfg{listen = listen}
+
+
+local function test_take_task_after_disconnect(test)
+    test:plan(1)
+    local driver = 'fifottl'
+    local tube = queue.create_tube('test_tube', driver,
+        {if_not_exists = true})
+    rawset(_G, 'queue', require('queue'))
+    tube:grant('guest', {call = true})
+    local task_id = tube:put('test_data')[1]
+    -- Now we have one task in a ready state
+
+    local connection = netbox.connect(listen)
+    local fiber_1 = fiber.create(function()
+        connection:call('queue.tube.test_tube:take')
+        connection:call('queue.tube.test_tube:take')
+    end)
+
+    -- This is not a best practice but we need to use the fiber.sleep()
+    -- (not fiber.yield()).
+    -- Expected results from a sleep() calling:
+    -- 1) Execute first connection:call('queue.tube.test_tube:take')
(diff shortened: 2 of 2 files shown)


t/110-take-task-after-reconnect.t
```diff
@@ -0,0 +1,47 @@
...
+
+    local connection = netbox.connect(listen)
+    local fiber_1 = fiber.create(function()
+            connection:call('queue.tube.test_tube:take')
+            connection:call('queue.tube.test_tube:take')
+        end)
+
+    fiber.sleep(0.1)
+    connection:close()
+    fiber.set_joinable(fiber_1, true)
```
olegrok: - You can't make joinable fiber that was created via "create". Only "new".
- I suppose you should call "set_joinable" right after fiber.new - without yields

See https://www.tarantool.io/en/doc/1.10/reference/reference_lua/fiber/#lua-function.fiber_object.set_joinable

And one nit: use `fiber1:set_joinable(true)` instead of `fiber.set_joinable(fiber1, true)`

LeonidVas: Hi, thank for the review. If I understood correctly, I can make a fiber joinable (that was created via create), but "the fiber could become ‘dead’ before fiber_object:set_joinable() takes effect". So, the fiber_1 will be hung on the second call of connection:call('queue.tube.test_tube:take') and don't be a 'dead'. Am I right?

olegrok: Yes, you are right, but let's look at this part again:

```lua
-- this a "main fiber"
local connection = netbox.connect(listen)
-- main fiber yields
local fiber_1 = fiber.create(function()
            -- fiber_1 yields
            connection:call('queue.tube.test_tube:take')
           -- fiber_1 yields
            connection:call('queue.tube.test_tube:take')
end)

-- main fiber yields
fiber.sleep(0.1)
-- There I don't know does main fiber yields on connection.close (I suppose yes, because it's a "I/O action")
-- but not 100% sure
connection:close()
fiber.set_joinable(fiber_1, true)
```

Hmm... I think it could be rewritten with channels is the following manner:

```lua
local ready_to_disconnect = fiber.channel()
local ready_to_take = fiber.channel()

fiber.create(function()
    connection:call('queue.tube.test_tube:take')
    ready_to_disconnect:put({})
    ready_to_take:get()
end)

ready_to_disconnect:get()
connection:close()
ready_to_take:put()
```
Please, consider it, I think it's much simpler and clear

LeonidVas: Hi. Thank you for the proposal. I am agree that the channels is more elegant decision, but it don't reproduce a bug because you has closed connection before a second call (connection:call('queue.tube.test_tube:take')).

queue/abstract.lua
```diff
@@ -352,6 +359,7 @@ function method._on_consumer_disconnect()
         box.space._queue_consumers:delete{ waiter[1], waiter[2] }
         local cond = conds[waiter[2]]
         if cond then
+            releasing_sessions[waiter[2]] = true
             cond:signal(waiter[2])
```
Totktonada: The reason why we do it here is to pass the fiber back to the fiber pool. I think fiber.cancel() perfectly fits this need. (NB: don't forget to free the condition variable.)

t/110-take-task-after-reconnect.t
```diff
@@ -0,0 +1,56 @@
...
+        end)
+
+    -- This is not a best practice but we need to use the fiber.sleep().
+    -- Expected results from a sleep() calling:
+    -- 1) Execute first connection:call('queue.tube.test_tube:take')
+    -- 2) Call the second connection:call('queue.tube.test_tube:take')
+    --  and to hang the fiber_1
+    -- 3) Start a fiber on the server side of connection which will execute
+    --  second queue.tube.test_tube:take call and hang because the queue
+    --  is empty
```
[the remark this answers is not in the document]
LeonidVas: It's about why we need to use a sleep instead of yield. And ok I'll update the bug description.

LeonidVas:
Updated

LeonidVas:
Now two implementations have been presented, and you need to choose one.

queue/abstract.lua
```diff
@@ -356,7 +356,9 @@ function method._on_consumer_disconnect()
         box.space._queue_consumers:delete{ waiter[1], waiter[2] }
         local cond = conds[waiter[2]]
         if cond then
-            cond:signal(waiter[2])
+            fiber_w = fiber.find(waiter[2])
+            fiber_w:cancel()
+            conds[waiter[2]] = nil
```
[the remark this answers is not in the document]
LeonidVas: fixed

Totktonada:
Leonid said that he met problems with approach with fiber.cancel(). The reason likely is caching of fiber conds (see compat.lua) and that `waiters` field of `struct fiber_cond` remains filled after canceling the fiber.

Let's stick with the original Leonid's variant.

Totktonada:
Pushed two small fixups.

Totktonada:
Squashed them (Leonid agreed).

PR #109, opened 2020-01-30
Validate indexes of queue space in driver.new() method

olmiik:
Currently for all queue implementations  `driver.new()` does not validate queue space, created by `driver.create_space()` method. However corrupted space can lead to unexpected behavior (for example, see #134),

This one adds check to ensure, that indexes, created by `driver.create_space()`, really exist.

### DIFF
--- a/queue/abstract/driver/fifo.lua
+++ b/queue/abstract/driver/fifo.lua
@@ -33,8 +33,21 @@ function tube.create_space(space_name, opts)
     return space
 end
 
+-- validate space of queue
+local function validate_space(space)
+    -- check indexes
+    if space.index.task_id == nil then
+        error('space does not have task_id index')
+    end
+    if space.index.status == nil then
+        error("space does not have status index")
+    end
+end
+
 -- start tube on space
 function tube.new(space, on_task_change)
+    validate_space(space)
+
     on_task_change = on_task_change or (function() end)
     local self = setmetatable({
         space          = space,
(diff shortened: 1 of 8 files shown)


queue/abstract/driver/fifo.lua
```diff
@@ -33,8 +33,21 @@ function tube.create_space(space_name, opts)
     return space
 end
 
+-- validate space of queue
+local function validate_space(space)
```
LeonidVas: I agree, WET practice can be used here.
But I suggest using something like:
"
local function validate_space(space, indexes)
	local indexes = {'task_id', 'status'}
	for _, index in pairs(indexes) do
		if space.index[index] == nil then
			error('space does not have ' .. index .. ' index')
		end
	end
"

t/010-fifo.t
```diff
@@ -292,6 +292,38 @@ test:test('if_not_exists test', function(test)
...
+    local fifo = require('queue.abstract.driver.fifo')
+    local space = fifo.create_space('corrupted_fifo_space', { engine = engine })
+    
+    local task_id = space.index.task_id
+    local status = space.index.status
+
+    test:test('task_id index does not exist', function(test)
+        test:plan(2)
+
+        space.index.task_id = nil
```
Totktonada: This is not proper use of the index lua api and it is not guaranteed to work in a future. OTOH, I don't see anything that would block usage of \<index_object\>:drop() and then re-creation of the space for a next test.

Totktonada:
@olmiik I want to kindly remind you about this PR. There are a couple of small comments. It would be good to resolve them and finally fix the problem for users.

LeonidVas:
Continued in PR https://github.com/tarantool/queue/pull/136

PR #119, opened 2020-03-26
travis-ci: fix publishing packages from tagged commits

### LINKED ISSUE #117: Tagged commits are not deployed
The change in 04c7cdb8bbd3492cd3eee7b4399985233a82ccc8 breaks deployment of tagged commits. See https://travis-ci.org/github/tarantool/queue/builds/656513720

Cast @printercu.

printercu:
Fixes #117

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ stages:
     # see:
     #   * https://github.com/tarantool/tarantool/issues/3745
     #   * https://github.com/travis-ci/travis-ci/issues/7780#issuecomment-302389370
-    if: (branch = master OR tag IS present) AND (env(PACKAGECLOUD_TOKEN) IS present)
+    # Also skip deploys for pull-requests, api and cron events.
+    if: (type = push) AND (branch = master OR tag IS present) AND (env(PACKAGECLOUD_TOKEN) IS present)
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -32,6 +32,8 @@ _deploy_to_packagecloud: &deploy_to_packagecloud
   dist: $OS/$DIST
   package_glob: build/*.{rpm,deb}
   skip_cleanup: true
+  on:
+    all_branches: true # it's restricted with the condition for deploy stage
```
Totktonada: Cron jobs should be explicitly disabled so. And also API request jobs. See [TRAVIS_EVENT_TYPE](https://docs.travis-ci.com/user/environment-variables/) variable.

Pull requests also should be excluded.

printercu: Aren't they filtered with `if: (branch = master OR tag IS present) AND (env(PACKAGECLOUD_TOKEN) IS present)`?

Totktonada: Nope. Not all at least.

printercu: Ok. Let's switch back to 2 deploys per repo.

Totktonada: I guess it would be possible to overcome the problem by checking `TRAVIS_EVENT_TYPE` in the snippet you cited. It still don't handle cron jobs and api request jobs (it'll try to deploy in them as I see); howevre we don't have them for now. PRs and commits into a fork will be handled by `env(PACKAGECLOUD_TOKEN) IS present` I guess (however I would prefer to set tarantool/queue repository for deployment explicitly).

So as I see both variants have the same behaviour and there is no reason to switch to '2 deploys per repo' way. I just proposed to make conditions a bit more explicit while we're here.

Totktonada:
> Please, prepend the commit message header with test: prefix. Add the issue number to the commit message.

Please, don't forget about this.

.travis.yml
Totktonada: Done.

PR #128, opened 2020-07-24
Fix release all tasks

### LINKED ISSUE #121: error in tube_release_all_tasks()
a question about the following function from abstract.lua:

```
local function tube_release_all_tasks(tube)
    local prefix = ('queue: [tube "%s"] '):format(tube.name)

    -- We lean on stable iterators in this function.
    -- https://github.com/tarantool/tarantool/issues/1796
    if not qc.check_version({1, 7, 5}) then
        log.error(prefix .. 'no stable iterator support: skip task releasing')
        log.error(prefix .. 'some tasks may stuck in taken state perpetually')
        log.error(prefix .. 'update tarantool to >= 1.7.5 or take the risk')
    end

    log.info(prefix .. 'releasing all taken task (may take a while)')
    local released = 0
    for _, task in tube.raw:tasks_by_state(state.TAKEN) do
        tube.raw:release(task[1])
        released = released + 1
    end
    log.info(prefix .. ('released %d tasks'):format(released))
end
```
this line `tube.raw:release(task[1])` leads to an error:
queue/abstract/driver/fifottl.lua:306: attempt to index local 'opts' (a nil value)
because the call to release() is missing the opts argument.

also, it seems that the version check branch is missing a return statement.

LeonidVas:
* Added return in case of tarantool version < 1.7.5
* Instead of nil, an empty table is used as "opts" argument of the driver "release" method. This is the policy of the module "queue": "Check opts in abstract.lua, instead to check inside the driver".
* Added a test for release all tasks on start.

Follow-up #66
Fixes #121

ChangeLog: fix the tasks releases on start for some drivers(utubettl, fifottl, limfifottl).

### DIFF
--- a/t/130-release-all-tasks-on-start.t
+++ b/t/130-release-all-tasks-on-start.t
@@ -1,0 +1,48 @@
+#!/usr/bin/env tarantool
+
+local os = require('os')
+local queue = require('queue')
+local tap = require('tap')
+local tnt = require('t.tnt')
+
+local test = tap.test('release all tasks on start')
+test:plan(1)
+
+---
+-- Accept gh-66, we need to release all taken tasks on start.
+-- Instead of tarantool reboot, we will additionally call queue.start()
+-- to simulate the reload of the module. This is not a clean enough,
+-- because fibers launched by the module are not cleaned up.
+-- See https://github.com/tarantool/queue/issues/66
+--
+-- All tricks in this test are done by professionals, don't try
+-- to repeat it yourself!!!
+local function check_release_tasks_on_start()
+    tnt.cfg()
+    -- The "fifottl" driver was choosen for check gh-121.
+    -- We cann't use opts == nil as argument for driver "release"
+    -- method. This is the policy of the module "queue" (check
+    -- opts in abstract.lua, instead to check inside the driver).
+    -- See https://github.com/tarantool/queue/issues/121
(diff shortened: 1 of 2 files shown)


t/130-release-all-tasks-on-start.t
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env tarantool
```
Totktonada, line 1: Please, add the executable bit for the file.

t/130-release-all-tasks-on-start.t
```diff
@@ -0,0 +1,48 @@
...
+test:plan(1)
+
+---
+-- Accept gh-66, we need to release all taken tasks on start.
+-- Instead of tarantool reboot, we will additionally call queue.start()
+-- to simulate the reload of the module. This is not a clean enough,
+-- because fibers launched by the module are not cleaned up.
+-- See https://github.com/tarantool/queue/issues/66
+--
+-- All tricks in this test are done by professionals, don't try
```
rybakit, line 18: 🤣

t/130-release-all-tasks-on-start.t
LeonidVas: Done.

LeonidVas:
> Please, add an entry for changelog to the PR.
Done.

PR #129, opened 2020-08-03
Add driver API check

### LINKED ISSUE #126: Fail gracefully at lack of tasks_by_state() driver API method
Driver API was changed with the new requirement in 75c47082b9a5e9fe66cb0a8d1579fbf2a63637cb, which was released in 1.0.6: the new method `tasks_by_state()` must be defined in a queue driver. We unable to relax this requirement without potentially loosing of task processing correctness (see #66).

All built-in drivers were updated, but there are external ones and we should give developers clear notification how to update a driver to support queue-1.0.6. Let's give clear error message in case of lack of the method and update the [release notes for 1.0.6](https://github.com/tarantool/queue/releases/tag/1.0.6) with upgrade path and simple example.

LeonidVas:
In addition to the core queue drivers, customer drivers are exists. In the commit a check for a driver API implementation was added. Now, the consumer will be informed about the missing methods in the driver implementation.

Closes #126

ChangeLog: added notification about missing methods in the driver implementation.
ReleaseNotes: the "tasks_by_state" method was added to the driver API. See example implementation in one of the core drivers.

### DIFF
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -308,6 +308,34 @@ end
 -- methods
 local method = {}
 
+-- List of required driver methods.
+local required_driver_methods = {
+    'normalize_task',
+    'put',
+    'take',
+    'delete',
+    'release',
+    'bury',
+    'kick',
+    'peek',
+    'touch',
+    'truncate',
+    'tasks_by_state'
+}
+
+-- gh-126 Check the driver API.
+local function check_driver_api(tube_impl, tube_type)
+    for _, v in pairs(required_driver_methods) do
+        if tube_impl[v] == nil then
+            error(string.format('The "%s" driver does not have an ' ..
+                                'implementation of method "%s".', tube_type, v))
+        end
+    end
+end
+
+-- Cache of already verified drivers.
+local checked_drivers = {}
+
 local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
     opts = opts or {}
     local self
@@ -362,6 +390,12 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
     }, {
         __index = tube
     })
+
+    if checked_drivers[tube_type] == nil then
+        check_driver_api(self.raw, tube_type)
+        checked_drivers[tube_type] = true
+    end
+
     self:on_task_change(opts.on_task_change)
     queue.tube[tube_name] = self
 


queue/abstract.lua
```diff
@@ -75,6 +75,12 @@ local function tube_release_all_tasks(tube)
         return
     end
 
+    -- Custom drivers may not have an implementation of
+    -- the "tasks_by_state" method.
+    if tube.raw['tasks_by_state'] == nil then
+        return
+    end
```
Totktonada, line 82: I'm against allowing a driver to leave the method non-implemented. queue module is unable to work correctly without this method.

If it would about, say, statistics and queue module would able to work correctly with reduced functionality, then it would be okay to make the method optional. But it is not our case.

I vote to give a meaningful error and keep the method be required. I hope pushing developers to do small extra work to support the new queue release is worthful considering that the result will be correctness of task states.

queue/abstract.lua
```diff
@@ -368,6 +407,12 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
     }, {
         __index = tube
     })
+
+    if checked_drivers[tube_type] == nil then
+        check_driver_api(self.raw, tube_type)
+        checked_drivers[tube_type] = true
+    end
```
Totktonada, line 414: How about doing this check inside `register_driver()` function? It already contains check for presence of a couple of driver functions.

I see, built-in drivers are registered using just `require()`, but nothing stops us from changing it to explicit calls of `queue.register_driver()`.

Okay, I see, there is another problem: two implementations of `register_driver()`, but I think we can just keep one in `init.lua` and zap one in `abstract.lua` with a comment that it is defined in `init.lua`.

BTW, it takes the question above about hot reload off the table.

Totktonada: Hm, I see that you already zapped `register_driver()` from `abstract.lua` within PR #130, but you still don't use the method to verify driver functions and postpone the check until a tube (re)creation. I don't see a reason for this.

queue/abstract.lua
```diff
@@ -314,6 +314,45 @@ end
...
+        if tube_impl[v] == nil then
+            log.warn('The "' .. tube_type ..'" driver doesn\'t have an'
+                     .. ' implementation of method "' .. v ..'". Some queue'
+                     .. ' functionality may not work properly.')
+        end
+    end
+end
+
+-- Cache of already verified drivers.
+local checked_drivers = {}
```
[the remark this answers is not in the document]
LeonidVas: > How about hot reload? It seems we'll skip the check if a driver will be reloaded and it does not look fully correct behaviour.
> 
> In fact it is not clear how to correctly reload a driver, but I think it is good to write new code with hot reload possibility in a mind.

IMHO, we can't do a clean reload of some drivers now, and our plans don't include such functionality. In PR #130 I disabled hot reloading of drivers.

queue/abstract.lua
LeonidVas: IMHO, it is more friendly to let users continue to work as they did, if possible. But ok, I will move it to required_driver_methods.

queue/abstract.lua
LeonidVas: > Hm, I see that you already zapped `register_driver()` from `abstract.lua` within PR #130, but you still don't use the method to verify driver functions and postpone the check until a tube (re)creation. I don't see a reason for this.

How? I can't check the driver methods before creating it. In `register_driver()` I can only check the `tube` methods.

queue/abstract.lua
```diff
@@ -314,6 +314,45 @@ end
...
+local new_driver_methods = {
+    'tasks_by_state'
+}
+
+-- gh-126 Check the driver API.
+local function check_driver_api(tube_impl, tube_type)
+    for _, v in pairs(required_driver_methods) do
+        if tube_impl[v] == nil then
+            error('The "' .. tube_type ..'" driver does not have an'
+                  .. ' implementation of method "' .. v ..'".')
```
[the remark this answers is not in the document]
LeonidVas: Fixed

queue/abstract.lua
LeonidVas: Updated

queue/abstract.lua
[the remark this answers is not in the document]
Totktonada: I don't insist, okay.

queue/abstract.lua
Totktonada: I see it now: `register_driver()` checks driver module functions, while here you add checks for driver instance methods.

queue/abstract.lua
Totktonada: Thanks!

queue/abstract.lua
```diff
@@ -308,6 +308,33 @@ end
 -- methods
 local method = {}
 
+-- List of required driver methods.
+local required_driver_methods = {
```
Totktonada, line 312: `'touch'` is forgotten.

LeonidVas: Fixed.

PR #130, opened 2020-08-03
Update "register_driver"

LeonidVas:
Changes of patchset:
* Double definition of the "register_driver" method has been removed
* Loading core drivers of the queue has been moved from abstract.lua to init.lua
* Added check for driver overriding
* Added description of the "register_driver" method
* Driver API description has been fixed
* Added description of the "touch" driver API method

ChangeLog:
* Check for driver overriding added. Previously, such check existed only for the core drivers in the case of a delayed start (when the queue module tried to initialize before the box configuration.).
* Description of the "register_driver" method was added.
* Description of the "touch" driver API method was added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,13 +614,13 @@ Registering a custom driver
 Driver class must implement the following API:
 
 1. `new` (constructs an instance of a driver), takes:
-  * the space object, in which the driver must store its tasks
-  * a callback to notify the main queue framework on a task state change
+    * the space object, in which the driver must store its tasks
+    * a callback to notify the main queue framework on a task state change
   (`on_task_change`)
-  * options of the queue (a Lua table)
+    * options of the queue (a Lua table)
 1. `create_space` - creates the supporting space. The arguments are:
-  * space name
-  * space options
+    * space name
+    * space options
 
 To sum up, when the user creates a new queue, the queue framework
 passes the request to the driver, asking it to create a space to
--- a/README.md
+++ b/README.md
@@ -601,6 +601,14 @@ notify the framework about the change.
 I.e. for normal operation, even errors during normal operation, there
 should be no exceptions.
 
+Registering a custom driver
+
+`register_driver(driver_name, tube_ctr)` - queue method is used to register
+  a custom driver. The arguments are:
+  * driver_name: unique driver name. Must be different from the core drivers
+  names.
+  * tube_ctr: implementation of tube control methods("create_space" and "new").
+
 ## Driver API
 
 Driver class must implement the following API:
--- a/README.md
+++ b/README.md
@@ -646,6 +646,8 @@ Returns the original task with a state changed to 'done'
 * `tube:bury(task_id)` - buries a task
 * `tube:kick(count)` - digs out `count` tasks
 * `tube:peek(task_id)` - return the task state by ID
+* `tube:touch(task_id, delta)` - increases `ttr` and `ttl` of the task by delta
+seconds. If queue does not support `ttr`, error will be thrown. Returns the task
 * `tube:tasks_by_state(task_state)` - return the iterator to tasks in a certain state
 * `tube:truncate()` - delete all tasks from the tube. Note that `tube:truncate`
 must be called only by the user who created this tube (has space ownership) OR
--- a/queue/init.lua
+++ b/queue/init.lua
@@ -1,5 +1,14 @@
 local queue = nil
 
+-- load all core drivers
+local core_drivers = {
+    fifo        = require('queue.abstract.driver.fifo'),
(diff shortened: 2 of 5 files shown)


queue/init.lua
```diff
@@ -6,6 +6,9 @@ local function register_driver(driver_name, tube_ctr)
         error('tube control methods must contain functions "create_space"'
               .. ' and "new"')
     end
+    if queue.driver[driver_name] then
+        error(('overriding registered driver "%s"'):format(driver_name))
+    end
```
Totktonada, line 11: So before box.cfg() we'll not receive an error on `register_driver()` on attempt to override a built-in dirver (it'll be postponed until box.cfg() call). But we'll get the error when will try the same after box.cfg().

For me it looks a bit lopsided and I would see whether moving `queue.driver` table from `abstract.lua` to `init.lua` may help. It would also remove 'overriding default driver' check from the box.cfg wrapper, which should be done right in `register_driver()`, IMHO.

t/140-register-driver-after-cfg.t
```diff
@@ -35,6 +35,11 @@ local function check_driver_register()
     end
 
     test:ok(check_standart_drivers, 'standard drivers are defined')
+
+    local res, err = pcall(queue.register_driver, 'mock', mock_tube)
+    local check = not res and
+        string.match(err, 'overriding registered driver') ~= nil
+    test:ok(check, 'check a driver override failure')
```
Totktonada, line 42: Maybe add the same test to `001-tube-init.t` to verify the same situation, when `register_driver()` is called before `box.cfg()`?

README.md
```diff
@@ -601,6 +601,14 @@ notify the framework about the change.
 I.e. for normal operation, even errors during normal operation, there
 should be no exceptions.
 
+Registering a custom driver
+
+`register_driver(driver_name, tube_ctr)` - queue method is used to register
+  a custom driver. The argumetns are:
```
Totktonada, line 607: Typo: argumetns.

queue/init.lua
LeonidVas: Updated.

t/140-register-driver-after-cfg.t
LeonidVas: Added.

README.md
LeonidVas: Fixed.

Totktonada (approved):
Thank you!

LGTM, but minor comments:

Regarding *'refactoring: update definition of "register_driver" method'*. AFAIU the commit does not change a behaviour, but it adds a test. It is not clear on the first glance, whether something is fixed or not. I would clarify this point.

Regarding *'refactoring: move loading core drivers to init.lua'*. It seems it is not just refactoring, because it actually changes a behaviour: the check re name clash against built-in drivers now occurs at `require('queue')` moment despite whether box is confgured or not. I don't know, how to better reflect this fact: remove 'refactoring' prefix or just note the difference in the behaviour in the commit message. I tends to interpret 'refactoring' as 'no actual behaviour changes', but maybe 'there are behaviour changes, which aren't the main reason of changes' is okay too (not sure).

Regarding changelog entries:

> Check for driver overriding added.

It was here before, but was missed in some cases. I would try to give exact case(s), where user-visible behaviour is changed. Please, try to describe it in a more explicit way, but feel free to ignore this comment if the description will became vague (sometimes it is hard to give an idea in tight words).

No need to re-review with me.

I guess it is safe enough to include into the upcoming release (whose main purpose is to fix regressions introduced in 1.0.6 and don't break anything else). Not strictly sure, though.

LeonidVas:
> Thank you!
> 
> LGTM, but minor comments:
> 
> Regarding _'refactoring: update definition of "register_driver" method'_. AFAIU the commit does not change a behaviour, but it adds a test. It is not clear on the first glance, whether something is fixed or not. I would clarify this point.

Tests were added to increase the tests coverage, so as not to break something during refactoring.
 
> 
> Regarding _'refactoring: move loading core drivers to init.lua'_. It seems it is not just refactoring, because it actually changes a behaviour: the check re name clash against built-in drivers now occurs at `require('queue')` moment despite whether box is confgured or not. I don't know, how to better reflect this fact: remove 'refactoring' prefix or just note the difference in the behaviour in the commit message. I tends to interpret 'refactoring' as 'no actual behaviour changes', but maybe 'there are behaviour changes, which aren't the main reason of changes' is okay too (not sure).
> 

'refactoring' prefix has been removed.

LeonidVas:
> It was here before, but was missed in some cases. I would try to give exact case(s), where user-visible behaviour is changed. Please, try to describe it in a more explicit way, but feel free to ignore this comment if the description will became vague (sometimes it is hard to give an idea in tight words).

ChangeLog updated.

PR #136, opened 2020-10-22
Add validation ot the space indexes

### LINKED ISSUE #62: tube:create() should check that a space with this name exists
If a space with the same name as tube name exists, queue module does not check that this space schema has structure which is different from the desired one, and goes on (if_not_exists clause is set to true).
This is error prone, since later on such queue will malfunction.

Check that the space used for queue actually is created with queue:create and complain if not.

LeonidVas:
* For all queue drivers, the "new" method now checks for the presence of indexes created by the "create_space" method.
* If a space is creating with the "if_not_exist = true" option and a space with such name already exists (and this is not a tube of the required type), a conflict will be detected and an error will be thrown. This is necessary because we don't want to create additional indexes for the existing space and repurpose it into the "tube".

Closes #62

@ChangeLog Added space schema validation to all core drivers when creating a tube object. If a space is creating with the "if_not_exist = true" option and a space with such name already exists (and this is not a tube of the required type), a conflict will be detected and an error will be thrown.

### DIFF
--- a/queue/abstract/driver/fifo.lua
+++ b/queue/abstract/driver/fifo.lua
@@ -6,6 +6,18 @@ local str_type = require('queue.compat').str_type
 local tube = {}
 local method = {}
 
+-- validate space of queue
+local function validate_space(space)
+    -- check indexes
+    local indexes = {'task_id', 'status'}
+    for _, index in pairs(indexes) do
+        if space.index[index] == nil then
+            error(string.format('space "%s" does not have "%s" index',
+                space.name, index))
+        end
+    end
+end
+
 -- create space
 function tube.create_space(space_name, opts)
     local space_opts         = {}
@@ -35,6 +47,8 @@ end
 
 -- start tube on space
 function tube.new(space, on_task_change)
+    validate_space(space)
+
     on_task_change = on_task_change or (function() end)
     local self = setmetatable({
         space          = space,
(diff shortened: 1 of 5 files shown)


queue/abstract/driver/fifo.lua
```diff
@@ -23,25 +23,30 @@ function tube.create_space(space_name, opts)
...
-    })
-    space:create_index('status',  {
-        type = 'tree',
-        parts = {2, str_type(), 1, num_type()},
-        if_not_exists = if_not_exists
-    })
+    local space = box.space[space_name]
+    if if_not_exists and space then
+        -- Validate the existing space.
+        validate_space(box.space[space_name])
```
Totktonada, line 36: We can reduce indentation level (and diff too) if we'll add `return space` here. Up to you.

Totktonada (approved):
I like the idea to offer extra safely over mistakes around space / tube names. I don't see any problems in the implementation. LGTM.

queue/abstract/driver/fifo.lua
```diff
@@ -6,35 +6,54 @@ local str_type = require('queue.compat').str_type
...
```
Totktonada, line 36: Raw idea: should not we verify that `box.space._queue` contains the relevant entry when the requested space exists and `if_not_exists` is positive? (We can do it for all drivers in `create_tube()` in abstract.lua.)

queue/abstract/driver/fifo.lua
LeonidVas: Yep. Applied.

queue/abstract/driver/fifo.lua
LeonidVas: No. Now a tube space validation and the "_queue" space are located at different levels of abstraction (driver vs abstract).

Totktonada: Okay.

PR #139, opened 2020-11-08
Introduce "shared" sessions

### LINKED ISSUE #85: Cannot :ack task after implicit net.box reconnect

LeonidVas:
Before the patchset a connection to server was synonym of the queue session. Now the session has a unique UUID (returned by the "queue.identify()" method), and one session can have many connections. To connect to an existing session, call "queue.identify(uuid)" with the previously obtained UUID.
Also, for the queue the `ttr` option has been added.
`ttr` in seconds - the time after which, if there is no active connection in the session, it will be released with all its tasks.

Closes #85 

@ChangeLog:
The "shared" sessions was added to the queue. Previously, a connection to server was synonym of the queue session. Now the session has a unique UUID (returned by the "queue.identify()" method), and one session can have many connections. To connect to an existing session, call "queue.identify(uuid)" with the previously obtained UUID.
Also, for the queue the `ttr` option has been added.
`ttr` in seconds - the time after which, if there is no active connection in the session, it will be released with all its tasks.


queue/abstract.lua
```diff
@@ -547,19 +609,27 @@ function method.start()
     end
 
     local _taken = box.space._queue_taken
+
+    -- Format of the temporary space has been changed.
+    -- So recreate it if needed.
+    if _taken ~= nil and _taken:format()[1]['type'] == 'unsigned' then
```
LeonidVas, line 662: Should be implemented some other way because it will cause a problem when downgrading to a previous version.

queue/abstract.lua
```diff
@@ -393,17 +413,47 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
     return self
 end
 
+--- Release all session tasks if the session has no actual connections.
+local function cleanup_session_tasks(queue_uuid)
+    local sesion_ids = box.space._queue_session_ids.index.queue_uuid:select(
```
Totktonada: Typo: sesion → session.

queue/abstract.lua
```diff
@@ -574,6 +586,24 @@ local function identification_init()
     end
 end
 
+--- Create an expiration fiber to cleanup expired sessions.
+local function create_expiration_fiber()
+    local exp_fiber = fiber.new(function()
```
Totktonada, line 591: Just noted: fiber.new() is supported since 1.10.0-27-g48cad35eb.

queue/abstract.lua
```diff
@@ -574,6 +586,24 @@ local function identification_init()
...
+        while true do
+            local cur_time = time()
+            for quuid, exp_time in pairs(inactive_sessions) do
+                if cur_time >= exp_time then
+                    cleanup_session_tasks(quuid)
```
Totktonada, line 596: Switching to read-only will kill the fiber (and moving back to read-write will not run it again). We can add a guardian fiber, which will restart the expiration fiber (like expirationd does). `pcall()` should be sufficient too.

queue/abstract.lua
```diff
@@ -495,6 +532,31 @@ function method.create_tube(tube_name, tube_type, opts)
     return self
 end
 
+--- Create everything that's needed to work with "shared" sessions.
+local function identification_init()
```
Totktonada, line 565: As raw idea: will the code looks better, if we'll move all session management stuff to its own file like queue/session.lua?

README.md
```diff
@@ -25,6 +25,7 @@ align="right">
 * [Using the queue module](#using-the-queue-module)
   * [Initialization](#initialization)
   * [Creating a new queue](#creating-a-new-queue)
+  * [Set queue settings](#set-queue-settings)
```
Totktonada, line 28: We usually mark relevant commits that follows `Closes #xxx` as `Follows up #xxx` rather than `Part of #xxx`.

queue/abstract.lua
```diff
@@ -393,17 +413,47 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
...
 end
 
+--- Release all session tasks if the session has no actual connections.
+local function cleanup_session_tasks(queue_uuid)
+    local sesion_ids = box.space._queue_session_ids.index.queue_uuid:select(
+        queue_uuid,
+        {limit = 1})[1]
+    if sesion_ids ~= nil then
+        return
+    end
```
Totktonada: It is a bit tricky, I would leave a comment here. I mean, I looked to the code a few minutes until I got the idea.

queue/abstract.lua
```diff
@@ -613,6 +685,38 @@ local function build_stats(space)
...
+--- Identificate the session.
+-- If "queue_uuid" == nil - create a new UUID for the session
+-- and return it, otherwise connect to the existing one.
+function method.identificate(queue_uuid)
+    local queue_session_ids = box.space._queue_session_ids
+    local sid = session.id()
+    local session_ids = queue_session_ids.index.box_sid:get(sid)
+    local quuid = session_ids and session_ids[2]
+
+    if queue_uuid == nil and quuid == nil then
```
Totktonada, line 744: Can we name those UUIDs somehow like `new_uuid` and `old_uuid`? It is hard to me to understand what is what when I look into some hunk within the function: the `queue_uuid` looks quite same to the `quuid`.

queue/abstract.lua
Totktonada: More I look into the code, more promising the idea looks for me. When I'm looking into queue code, I want to concentrate on the data layout, ordering guarantees, wakeups and other queue logic. Not on session management.

LeonidVas:
The [benchmark](https://github.com/tarantool/queue/blob/1d8f708e12e4b36301e325cccd46183fae89c609/t/benchmark/multi_consumer_work.lua) was run before and after adding "shared" sessions.
The command that was used to run the test: `rm -rf *.snap *.xlog && taskset 3 tarantool t/benchmark/multi_consumer_work.lua`
For each variant of the parameters, ten runs of tests were carried out, and after the result was averaged.
Result format: (time it takes to fill the queue / time it takes to confirm the tasks) in microseconds.

Master:

| consumers<br>batches  | 1                 | 100                   | 10000                 |
|---------------------- |----------------   |--------------------   |--------------------   |
| 1                     | 102 / 393         | 5479 / 10681          | 65914 / 322383        |
| 100                   | 7272 / 9107       | 49043 / 224925        | 4962018 / 33961217    |
| 10000                 | 41834 / 220479    | 3972183 / 25542455    |                       |

With "shared" sessions:

| consumers<br>batches  | 1                 | 100                   | 10000                 |
|---------------------- |----------------   |--------------------   |--------------------   |
| 1                     | 123 / 446         | 2776 / 12724          | 64437 / 509038        |
| 100                   | 7900 / 8411       | 55467 / 274383        | 4899408 / 35677578    |
| 10000                 | 54831 / 266131    | 4259797 / 30703081    |                       |

queue/abstract.lua
LeonidVas: Done.

queue/abstract.lua
LeonidVas: Done.

queue/abstract.lua
LeonidVas: Done.

queue/abstract.lua
LeonidVas: Done.

queue/abstract.lua
LeonidVas: Done.

README.md
LeonidVas: Done.

queue/abstract.lua
LeonidVas: Done.

queue/abstract.lua
LeonidVas: Done.

queue/abstract.lua
```diff
@@ -613,6 +732,47 @@ local function build_stats(space)
     return stats
 end
 
+--- Identificate the session.
+-- If "queue_uuid" == nil - create a new UUID for the session
+-- and return it, otherwise connect to the existing one.
+function method.identificate(queue_uuid)
```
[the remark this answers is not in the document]
LeonidVas: Changed to `identify`.

LeonidVas:
After refactoring.

Master:

| consumers<br>batches  | 200               | 400                   |
|---------------------- |----------------   |--------------------   |
| 200                   | 132887 / 659421   | 268149 / 1490263      |
| 400                   | 247721 / 1538678  | 480667 / 3362063      |

With "shared" sessions:

| consumers<br>batches  | 200               | 400                   |
|---------------------- |----------------   |--------------------   |
| 200                   | 136281 / 620819   | 259882 / 1426101      |
| 400                   | 249045 / 1361905  | 490443 / 2613279      |

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,192 @@
...
+    local session_ids = queue_session_ids.index.box_sid:get(conn_id)
+    local cur_uuid = session_ids and session_ids[2]
+
+    if session_uuid == nil and cur_uuid == nil then
+        -- Generate new UUID for the session.
+        cur_uuid = uuid.bin()
+        queue_session_ids:insert{conn_id, cur_uuid}
+    elseif session_uuid ~= nil then
+        --Validate UUID.
+        uuid.frombin(session_uuid)
```
olegrok, line 100: Not sure it's good to show such message for users
```
tarantool> uuid.frombin(a)
---
- error: 'builtin/uuid.lua:73: frombin(bin, [byteorder])'
...
```

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,192 @@
...
+    local queue_session_ids = box.space._queue_session_ids
+    local session_ids = queue_session_ids.index.box_sid:get(conn_id)
+    local cur_uuid = session_ids and session_ids[2]
+
+    if session_uuid == nil and cur_uuid == nil then
+        -- Generate new UUID for the session.
+        cur_uuid = uuid.bin()
+        queue_session_ids:insert{conn_id, cur_uuid}
+    elseif session_uuid ~= nil then
+        --Validate UUID.
```
olegrok, line 99: Usually we put a space between `--` and text.

queue/abstract.lua
```diff
@@ -523,11 +535,11 @@ function method.start()
 
     local _cons = box.space._queue_consumers
     if _cons == nil then
-        -- session, fid, tube, time
+        -- connection, fid, tube, time
         _cons = box.schema.create_space('_queue_consumers', {
             temporary = true,
             format = {
-                {name = 'session_id', type = num_type()},
+                {name = 'connection_id', type = num_type()},
```
olegrok, line 542: I'm not sure we can do that. Have you tried to upgrade queue created before this patch and after?

After format won't changed because `box.space._queue_consumers` is not nil.
```lua
tarantool> box.schema.space.create('tmp', {temporary = true})
---
- engine: memtx
  before_replace: 'function: 0x0114821f38'
  field_count: 0
  is_sync: false
  ck_constraint: []
  on_replace: 'function: 0x0114821f00'
  temporary: true
  index: []
  is_local: false
  enabled: false
  name: tmp
  id: 539
- created
...

tarantool> box.space.tmp
---
- engine: memtx
  before_replace: 'function: 0x0114821f38'
  field_count: 0
  is_sync: false
  ck_constraint: []
  on_replace: 'function: 0x0114821f00'
  temporary: true
  index: []
  is_local: false
  enabled: false
  name: tmp
  id: 539
...
-- restart instance here. It will be cleared by physically it still exists
box.space.tmp
---
- engine: memtx
  before_replace: 'function: 0x01084723e0'
  field_count: 0
  is_sync: false
  ck_constraint: []
  on_replace: 'function: 0x0108472220'
  temporary: true
  index: []
  is_local: false
  enabled: false
  name: tmp
  id: 539
...

```

queue/abstract.lua
```diff
@@ -393,45 +388,46 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
     return self
 end
 
+--- Release all session tasks.
+local function release_session_tasks(session_uuid)
+    local taken_tasks = box.space._queue_taken_2.index.uuid:select(session_uuid)
+
+    for _, task in pairs(taken_tasks) do
```
olegrok, line 395: Since `pairs `can't be JIT-compiled[1] it's preferable to use `ipairs` here

[1] http://wiki.luajit.org/NYI#libraries_base-library

queue/abstract.lua
```diff
@@ -393,45 +388,46 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
     return self
 end
 
+--- Release all session tasks.
+local function release_session_tasks(session_uuid)
+    local taken_tasks = box.space._queue_taken_2.index.uuid:select(session_uuid)
```
olegrok, line 393: Here you use key wrapped in parenthesis but below you use delete with braces. I think it should be unified.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,192 @@
...
+        session_uuid = session_ids[2]
+    else
+        session_uuid = queue_session.identify(connection_id)
+    end
+
+    return session_uuid
+end
+
+local function cleanup_inactive_sessions()
+local cur_time = util.time()
```
olegrok, line 57: Seems you've missed indent here

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,192 @@
...
+-- release its tasks if necessary.
+local function remove_connection(conn_id)
+    local queue_session_ids = box.space._queue_session_ids
+    local session_uuid = queue_session.get_uuid(conn_id)
+
+    queue_session_ids:delete{conn_id}
+    local session_ids = queue_session_ids.index.uuid:select(session_uuid,
+        {limit = 1})[1]
+
+    -- If a queue session don't have any active connections it should be
```
olegrok, line 143: `don't` -> `doesn't`

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,192 @@
...
+
+    for key, val in pairs(opts) do
+        if key == 'ttr' then
+            if val < 0 then
+                error('Invalid value of ttr: ' .. val)
+            end
+        else
+            error('Unknown option ' .. key)
+        end
+    end
```
olegrok, line 166: I believe it should be moved to "validate_settings" function

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,192 @@
...
+        { if_not_exists = true })
+end
+
+local function start()
+    identification_init()
+    queue_session.expiration_fiber = create_expiration_fiber()
+end
+
+-- methods
+local method = {
```
olegrok, line 182: May be `methods`?

README.md
```diff
@@ -349,6 +361,32 @@ created.
...
+## Set queue settings
+
+```lua
+queue.set_settings({options})
+```
+
+Set queue settings.  
+In case of failure, an error will be thrown.
+Available `options`:
+`ttr` - time to release in seconds. The time after which, if there is no active
```
olegrok, line 373: I think it should be a list. Currently it looks quite strange:
image

queue/abstract.lua
```diff
@@ -125,25 +107,34 @@ function tube.take(self, timeout)
     end
 end
 
+--- Checks is the task has been taken in a current session or in a session
+-- with session uuid = session_uuid.
+-- Throw an error on failure.
+local function check_task_on_taken(tube_id, task_id, session_uuid)
+    local _taken = box.space._queue_taken_2.index.task:get{tube_id, task_id}
+
+    if _taken == nil or ((session_uuid ~= nil and _taken[4] ~= session_uuid) and
```
olegrok, line 116: I suggest to put "4" to constant.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,192 @@
...
+end
+
+--- Remove a connection from the list of active connections and
+-- release its tasks if necessary.
+local function remove_connection(conn_id)
+    local queue_session_ids = box.space._queue_session_ids
+    local session_uuid = queue_session.get_uuid(conn_id)
+
+    queue_session_ids:delete{conn_id}
+    local session_ids = queue_session_ids.index.uuid:select(session_uuid,
```
olegrok, line 140: AFAIK you have some performance test for this feature. Could you please check is it reasonable or not to use "count" instead of "select" here? And don't create Lua objects.

olegrok (commented):
Thanks for your patch. Consider several comments below. I'll give more detailed review a bit later.

Also I think we could push d8de8b9aaedff9af3e7377c2b5a56ee316569fb5 out of order to simplify patchset and reduce diff.

queue/abstract.lua
```diff
@@ -374,6 +374,28 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
     return self
 end
 
+--- Release all session tasks.
+local function release_session_tasks(connection_id)
+    while true do
+        task = box.space._queue_taken.index.pk:min{connection_id}
```
olegrok: This variable is not declared.

Also I think you could use `ipairs` + REQ iterator here.
Consider following diff:

```diff
diff --git a/queue/abstract.lua b/queue/abstract.lua
index d666751..8a53aa6 100644
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -376,12 +376,7 @@ end
 
 --- Release all session tasks.
 local function release_session_tasks(connection_id)
-    while true do
-        task = box.space._queue_taken.index.pk:min{connection_id}
-        if task == nil or task[1] ~= connection_id then
-            break
-        end
-
+    for _, task in box.space._queue_taken.index.pk:pairs({connection_id}, {iterator = box.index.REQ}) do
         local tube = box.space._queue.index.tube_id:get{task[2]}
         if tube == nil then
             log.error("Inconsistent queue state: tube %d not found", task[2])
```

I think it should be faster. Main problem here is stability of TREE iterators (https://github.com/tarantool/tarantool/commit/2160e4cbdb90caff0a3c766bd92003d1f7d67d7b), we could break compatibility with old versions. But I think we could workaround it separately (but anyway since we don't have 1.7 in our CI we don't have any guarantees it works just now).

queue/abstract.lua
```diff
@@ -481,6 +481,18 @@ function method.create_tube(tube_name, tube_type, opts)
     return self
```
olegrok, line 491: commit message: seems it should be `The patch adds` instead of `The patch added`

queue/abstract.lua
olegrok: I think this condition is overcomplicated.

consider a following way
```lua
if session_uuid == nil then
    session_uuid = session.get_uuid(connection.id())
end

if _taken == nil or  _taken[4] ~= session_uuid then
```

queue/abstract.lua
```diff
@@ -375,49 +389,45 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
...
     end
 end
 
 function method._on_consumer_disconnect()
-    local waiter, fb, task, tube, id
-    id = connection.id()
+    local conn_id = connection.id()
+
     -- wakeup all waiters
     while true do
```
olegrok, line 413: Seems this loop could be also rewritten with ipairs

queue/abstract/queue_session.lua
olegrok: No, I was wrong. Event count = 100 it work slower than select.
```lua
local uuid = require('uuid')
local clock = require('clock')

box.cfg{}

local s = box.schema.space.create('test')
s:create_index('pk')
s:create_index('uuid', {unique = false, parts = {2, 'string'}})

box.begin()
local uuid_bin = uuid.bin()
for i = 1, 1e2 do
    s:replace{i, uuid_bin}
end
box.commit()

local s_select = clock.time()
for _ = 1, 1e5 do
    s.index.uuid:select({uuid_bin}, {limit = 1})
end
print('select', clock.time() - s_select)

local s_count = clock.time()
for _ = 1, 1e5 do
    s.index.uuid:count({uuid_bin}, {limit = 1})
end
print('count', clock.time() - s_count)

-- select	0.098401069641113
-- count	0.32897591590881
```

README.md
```diff
@@ -245,24 +248,33 @@ Consumers may be simply waiting for tasks to be put in the queues.
...
+1. `session_uuid` - session UUID (string)
 1. `time` - the time when the client began to execute the task
 
+The `_queue_session_ids` temporary space contains a map: box session id to the
+session UUID.
+
+## Fields of the `_queue_session_ids` space
+
+1. `box_session_id` - connection id (numeric)
+2. `session_uuid` - session UUID (string)
```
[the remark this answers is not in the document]
LeonidVas: This is a good point. But AFAIU then we should have a fallback for versions < 2.4.

queue/abstract.lua
LeonidVas: The "name" is not used inside the queue module.

README.md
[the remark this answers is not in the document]
olegrok: Yes. Moreover we should consider a case when queue from 2.4- will be updated to 2.4+

queue/abstract.lua
LeonidVas: Thank you for the information. This code has been moved in "refactoring: move" session release tasks "to a separate function" and I don't see any reasons to change this logic within the current task. But if we will have benchmark that will demonstrate profit from such changes, they can be implemented by a separate PR.

queue/abstract/queue_session.lua
LeonidVas: What for? `set_settings` is a small function and validation is the only thing that it does.

queue/abstract.lua
olegrok: I think changing pairs to ipairs we could consider as refactoring. Small note in commit message is more than enough IMO.

queue/abstract/queue_session.lua
LeonidVas: The name is the same as in abstract.lua

olegrok: Ok

queue/abstract.lua
olegrok:
```
local clock = require('clock')

local t = {}

for i = 1, 1e8 do
    table.insert(t, i)
end

local start = clock.time()
for _ in pairs(t) do end
print('pairs ', clock.time() - start)

collectgarbage()
collectgarbage()

local start = clock.time()
for _ in ipairs(t) do end
print('ipairs ', clock.time() - start)

os.exit()
```

Results:
```
pairs 	0.36381793022156
ipairs 	0.071022033691406
```

queue/abstract.lua
LeonidVas: > I suggest to put "4" to constant.

Yes. All of these numbers are very inconvenient when reading or writing code. I plan to replace it with some kind of `enum`, as it was done for the `status`. But this will be done by a separate PR. And these changes are beyond the scope of this task.

queue/abstract.lua
Mons: Let's test it with `jit.off()`
```
$ tarantool bench1.lua
pairs 	1.0151600837708
ipairs 	2.2498149871826
```

queue/abstract.lua
LeonidVas: > I think this condition is overcomplicated.
> 
> consider a following way
> 
> ```lua
> if session_uuid == nil then
>     session_uuid = session.get_uuid(connection.id())
> end
> 
> if _taken == nil or  _taken[4] ~= session_uuid then
> ```

Such condition allow to avoid calling `session.get_uuid()` in case of `_taken == nil`. But I think you are right and the simple condition is preferable.

queue/abstract.lua
LeonidVas: > Also I think you could use `ipairs` + REQ iterator here.
> Consider following diff:

https://github.com/tarantool/queue/pull/139#discussion_r544166212

LeonidVas: > This variable is not declared.

Yes. You are right (I haven't checked intermediate patches with `luachek` and I missed this mistake).

queue/abstract.lua
igormunkin: Sorry, but this is syntetic bullshit that doesn't relate to the actual code. Lua code might fail to be compiled due to other NYIs within loop body or even particular data being processed (see tarantool/tarantool#4630). Do you have any benchmarks and trace dumps for queue (at least) or rather performance critical application using this module? If not, the let's stop the discussion right here.

By the way, Leonid arguments are more reasonable than unproven performance degradation.

olegrok: OK, let's stay it as is.

LeonidVas:
> Thanks for your patch. Consider several comments below. I'll give more detailed review a bit later.

Hi! Thanks for the review.

> Also I think we could push [d8de8b9](https://github.com/tarantool/queue/commit/d8de8b9aaedff9af3e7377c2b5a56ee316569fb5) out of order to simplify patchset and reduce diff.

No, it's needed fo https://github.com/tarantool/queue/pull/139/commits/a3d6f4ea51df9b49c21415373e0f6677458188ea

queue/abstract/queue_session.lua
LeonidVas: Updated.

queue/abstract/queue_session.lua
LeonidVas: Updated.

queue/abstract.lua
LeonidVas: Updated.

queue/abstract/queue_session.lua
LeonidVas: yep. Updated

queue/abstract/queue_session.lua
LeonidVas: updated.

README.md
LeonidVas: Done.

queue/abstract.lua
LeonidVas: Updated.

README.md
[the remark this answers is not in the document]
olegrok: Seems it should be investigated one day. But seems "bin" is more lightweight than fair "uuid"

```
tarantool> #mp.encode(uuid.bin())
---
- 17
...

tarantool> #mp.encode(uuid())
---
- 18
...
```

queue/abstract.lua
```diff
@@ -374,6 +374,28 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
...
 
+--- Release all session tasks.
+local function release_session_tasks(connection_id)
+    while true do
+        local task = box.space._queue_taken.index.pk:min{connection_id}
+        if task == nil or task[1] ~= connection_id then
+            break
+        end
+
+        local tube = box.space._queue.index.tube_id:get{task[2]}
```
olegrok: Raw idea 2:

Snippet 1:

```lua
box.cfg{}

local space = box.schema.space.create('test')
space:create_index('pk')
box.begin()
for i = 1, 1e6 do
    space:replace{i}
end
box.commit()

local start = clock.time()
for i = 1, 1e6 do
    space:get{i}
end
print('get', clock.time() - start)

space:drop()
box.snapshot()

os.exit()
```

Snippet 2
```lua
box.cfg{}

local space = box.schema.space.create('test')
space:create_index('pk')
box.begin()
for i = 1, 1e6 do
    space:replace{i}
end
box.commit()

local start = clock.time()
for i = 1, 1e6 do
    space.index[0]:count{i}
end
print('count', clock.time() - start)

space:drop()
box.snapshot()

os.exit()
```

Snippet 1:  `get	0.72974896430969`
Snippet 2: `count	0.31964302062988`

Maybe it's not critical for _queue space

queue/abstract.lua
```diff
@@ -375,49 +389,45 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
...
+            log.error("Inconsistent queue state: tube %d not found", task[1])
+            box.space._queue_taken_2.index.task:delete{task[1], task[2]}
         else
-            log.warn("Consumer %s disconnected, release task %s(%s)",
-                connection_id, task[3], tube[1])
-
-            tube_release_internal(queue.tube[tube[1]], task[3], nil,
-                                  connection_id)
+            log.warn("Session %s closed, release task %s(%s)",
+                uuid.frombin(session_uuid):str(), task[2], tube[1])
```
olegrok, line 402: This place could be optimized:
  - Even log level is "error" conversation from string-to-uuid-to-string will be performed.
  - Every time we allocate temporary uuid object.

Consider https://github.com/tarantool/tarantool/blob/e27745ad54c124e5f7a33eaaecec62ef0e9c58d1/src/lua/uuid.lua#L77-L87 we could implement function that uses preallocated buffer for uuid.

olegrok: For future, not in this patch

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,194 @@
...
+                queue_session._on_session_remove(session_uuid)
+            end
+            inactive_sessions[session_uuid] = nil
+        end
+    end
+end
+
+--- Create an expiration fiber to cleanup expired sessions.
+local function create_expiration_fiber()
+    local exp_fiber = fiber.create(function()
```
olegrok, line 70: I think we should add names for all fibers that we create.

fiber.self():name('queue.session_expiration_fiber')

t/benchmark/multi_consumer_work.lua
```diff
@@ -0,0 +1,90 @@
...
+    -- Wait for all tasks to be acked.
+    wait_all()
+    -- Complete the timing of task confirmation.
+    local complete_time = clock.proc64()
+
+    --Print the result in microseconds
+    print("Time it takes to fill the queue: " ..
+        tostring((start_ack_time - start_put_time) / 10^3))
+    print("Time it takes to confirm the tasks: " ..
+        tostring((complete_time - start_ack_time) / 10^3))
```
olegrok, line 81: I suggest to use (`%.3f`):format(...) for numbers.

t/benchmark/multi_consumer_work.lua
```diff
@@ -0,0 +1,90 @@
...
+
+box.cfg()
+
+local test_queue = queue.create_tube('test_queue', 'fifo', {temporary = true})
+
+local function prepare_consumers()
+    local consumers = {}
+
+    for i = 1, consumers_count do
+        consumers[i] = fiber.new(function()
```
olegrok, line 24: There is no significant difference between fiber.new and fiber.create. To be compatible with Tarantool 1.9 we could use fiber.create. "new" was introduced only in 1.10. But up to you.

t/benchmark/multi_consumer_work.lua
```diff
@@ -0,0 +1,90 @@
...
+            wait_count = wait_count + 1
+            -- Wait for all consumers to start.
+            barrier:wait()
+
+            -- Create test tasks.
+            for j = 1, batch_size do
+                test_queue:put('test, cons: ' .. i)
```
olegrok, line 31: Maybe it's not completely fair, because some time was spent for concatenation. Maybe you could create target string only once and then reuse it several times.

queue/abstract.lua
```diff
@@ -393,45 +388,46 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
...
+
 function method._on_consumer_disconnect()
-    local waiter, fb, task, tube, id
-    id = session.id()
+    local conn_id = connection.id()
+
     -- wakeup all waiters
     while true do
-        waiter = box.space._queue_consumers.index.pk:min{id}
+        waiter = box.space._queue_consumers.index.pk:min{conn_id}
```
olegrok, line 414: uninitalized variable

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,194 @@
...
+        while true do
+            if box.info.ro == false then
+                local status, err = pcall(cleanup_inactive_sessions)
+                if status == false then
+                    log.error('An error occured while cleanup the sessions: %s',
```
olegrok, line 75: typo: `occured` -> `occurred`

queue/abstract.lua
LeonidVas: Sorry, but I didn't understand you. Can you create a task, send a PR or write  an example here?

olegrok: ok

queue/abstract/queue_session.lua
LeonidVas: Done.

t/benchmark/multi_consumer_work.lua
LeonidVas: I think microsecond is enough accuracy and used (`% i`).

olegrok (approved):
Thanks for your patch, thanks for your fixes. LGTM.

t/benchmark/multi_consumer_work.lua
LeonidVas: I don't mind (in this case). Updated.

t/benchmark/multi_consumer_work.lua
LeonidVas: Updated.

queue/abstract.lua
LeonidVas: Yep. Updated.

queue/abstract/queue_session.lua
LeonidVas: Fixed.

queue/abstract.lua
```diff
@@ -125,6 +125,16 @@ function tube.take(self, timeout)
...
 
+--- Checks is the task has been taken in a session with
+-- connection id == "conn_id".
+-- Throw an error on failure.
+local function check_task_on_taken(tube_id, task_id, conn_id)
+    local _taken = box.space._queue_taken:get{conn_id, tube_id, task_id}
+    if _taken == nil then
+        error("Task was not taken in the session")
+    end
+end
```
Totktonada: Several nits, feel free to ignore.

1. I would not mix helpers with API function. I think it is better to move helpers at top.
2. Typo: 'Checks is' → 'if'.
3. 'Throw an error on failure' — it is unclear from the description, whether taken or not taken task is considered as failure.
4. `check_task_on_taken()` sounds strange. Variants: `check_task_is_taken()`, `check_task_is_owned()`, `check_task_ownership()`. Not sure here.
5. Since there is a usage of the function as a predicate, I would split it to `task_is_taken()` and `error_task_is_not_taken()` functions or kinda. Pcalling of a simple helper looks dubios for me. Alternative: `task_is_taken()` / `task_is_owned()`, which returns `is_taken, err`, so it may be placed directly to `assert()`.

queue/abstract.lua
```diff
@@ -152,16 +159,14 @@ function tube.touch(self, id, delta)
...
-    end
+    local conn_id = connection.id()
+    check_task_on_taken(self.tube_id, id, conn_id)
     local tube = box.space._queue:get{self.name}
     local space_name = tube[3]
 
     self:peek(id)
     -- delete task
-    box.space._queue_taken:delete{connection.id(), self.tube_id, id}
+    box.space._queue_taken:delete{conn_id, self.tube_id, id}
```
Totktonada: Nit: Does not look relevant to this commit.

queue/abstract.lua
```diff
@@ -129,8 +129,8 @@ end
 -- connection id == "conn_id".
 -- Throw an error on failure.
 local function check_task_on_taken(tube_id, task_id, conn_id)
```
Totktonada: AFAIU, the idea is to use {tube_id, task_id} to search a task instead of {conn_id, tube_id, task_id}, while both indices are unique. It would be great to leave some explanation, why it is faster.

queue/abstract/driver/fifottl.lua
```diff
@@ -107,7 +90,7 @@ local ttl_states    = { state.READY, state.BURIED }
 local ttr_state     = { state.TAKEN }
 
 local function fifottl_fiber_iteration(self, processed)
-    local now       = time()
+    local now       = util.time()
     local task      = nil
     local estimated = TIMEOUT_INFINITY
```
Totktonada, line 95: I would leave only one TIMEOUT_INFINITY constant, in the util module. This commit left TIMEOUT_INFINITY in fifottl.lua and utubettl.lua. It is inconsistent.

queue/util.lua
```diff
@@ -0,0 +1,41 @@
+local fiber = require('fiber')
+
+-- MAX_TIMEOUT == 100 years
+local MAX_TIMEOUT      = 365 * 86400 * 100
+-- Set to TIMEOUT_INFINITY
+-- instead returns time for next event
+local TIMEOUT_INFINITY = 18446744073709551615ULL
```
Totktonada, line 7: What is the difference?

queue/util.lua
```diff
@@ -0,0 +1,41 @@
...
+        return TIMEOUT_INFINITY
+    end
+    tm = 0ULL + tm * 1000000 + fiber.time64()
+    return tm
+end
+
+local util = {
+    max_timeout = MAX_TIMEOUT,
+    timeout_infinity = TIMEOUT_INFINITY
+}
```
Totktonada, line 33: They are still constants even despite that they are fields of a table. I would keep the upper case to highlight this fact.

queue/util.lua
```diff
@@ -0,0 +1,41 @@
...
+
+-- methods
+local method = {
+    time = time,
+    event_time = event_time
+}
+
+return setmetatable(util, { __index = method })
```
Totktonada, line 41: Hm. Is there any difference from the following?

```
{
    max_timeout = MAX_TIMEOUT,
    timeout_infinity = TIMEOUT_INFINITY,
    time = time,
    event_time = event_time,
}
```

queue/util.lua
```diff
@@ -0,0 +1,41 @@
+local fiber = require('fiber')
+
+-- MAX_TIMEOUT == 100 years
+local MAX_TIMEOUT      = 365 * 86400 * 100
+-- Set to TIMEOUT_INFINITY
+-- instead returns time for next event
+local TIMEOUT_INFINITY = 18446744073709551615ULL
+
+local function time(tm)
```
Totktonada, line 9: It transforms what to what? The same question for event_time().

Since you're in the context now, I would ask you to add a comment for those functions.

Sorry for nitpicking.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,195 @@
...
+                local status, err = pcall(cleanup_inactive_sessions)
+                if status == false then
+                    log.error('An error occurred while cleanup the sessions: %s',
+                        err)
+                end
+            end
+            fiber.sleep(1)
+        end
+    end)
+    exp_fiber:name('queue_expiration_fiber')
```
Totktonada, line 82: I would better set it in the first statement from inside the fiber rather than in an unclear moment after the fiber creating.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,195 @@
...
+        if key == 'ttr' then
+            if val < 0 then
+                error('Invalid value of ttr: ' .. val)
+            end
+        else
+            error('Unknown option ' .. key)
+        end
+    end
+
+    queue_session.settings = opts
```
Totktonada, line 171: If I'll do:

```lua
queue_session.set_settings({opt_1 = 'foo"})
queue_session.set_settings({opt_2 = 'bar"})
```

I would expect that the resulting values will be the following:

```lua
{
    opt_1 = 'foo',
    opt_2 = 'bar',
}
```

The same for the corresponding function in abstract.lua.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,195 @@
...
+    queue_session.expiration_fiber = create_expiration_fiber()
+end
+
+-- methods
+local method = {
+    get_uuid = get_uuid,
+    identify = identify,
+    on_session_remove = on_session_remove,
+    remove_connection = remove_connection,
+    set_settings = set_settings,
```
Totktonada, line 190: How about `cfg`, like in the `box`, `msgpack`, `yaml`, `json` built-in modules?

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,195 @@
...
+local inactive_sessions = {}
+
+--- Create everything that's needed to work with "shared" sessions.
+local function identification_init()
+    local queue_session_ids = box.space._queue_session_ids
+    if queue_session_ids == nil then
+        queue_session_ids = box.schema.create_space('_queue_session_ids', {
+            temporary = true,
+            format = {
+                {name = 'box_session_id', type = num_type()},
```
Totktonada, line 25: You call it 'connection_id' at least in the `get_uuid()` arguments. The module is abstracted out from box.session, so I would call it 'connection_id' uniformly.

The same for `box_sid` index name below.

queue/abstract.lua
```diff
@@ -7,7 +7,7 @@ local qc       = require('queue.compat')
 local num_type = qc.num_type
 local str_type = qc.str_type
 
-local session = box.session
+local connection = box.session
```
Totktonada, line 16: It looks as a 'connection' module that is a bit strange: we have no such server-side entity. How about defining a function to obtain a connection id, like so: `local connection_id = box.session.id`? Not sure here.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,195 @@
...
+end
+
+local function start()
+    identification_init()
+    queue_session.expiration_fiber = create_expiration_fiber()
+end
+
+-- methods
+local method = {
+    get_uuid = get_uuid,
```
Totktonada, line 186: We usually omit `get_` for getters. How about `connection_uuid()`? It'll also more descriptive and more uniform if we'll rename `remove_connection()` to `connection_remove()` (so the `connection_` will become a prefix).

What also disquets me a bit: there is no `add_connection()` or `connection_add()`, but there are `get_uuid()` and `identify()`. Maybe I'm too tired, but it looks hard to understand the difference between the two latter functions.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,195 @@
...
+    identify = identify,
+    on_session_remove = on_session_remove,
+    remove_connection = remove_connection,
+    set_settings = set_settings,
+    session_grant = session_grant,
```
Totktonada, line 191: It gives the ability to magane sessions from given user, right? How about just `grant()`?

queue/abstract.lua
```diff
@@ -640,6 +664,13 @@ local function build_stats(space)
     return stats
 end
 
+--- identify the session.
```
Totktonada, line 667: This description is a bit confusing for me. I would say something like 'Associate a current connection with given session'.

queue/abstract.lua
```diff
@@ -375,49 +389,45 @@ local function make_self(driver, space, tube_name, tube_type, tube_id, opts)
...
 
 function method._on_consumer_disconnect()
-    local waiter, fb, task, tube, id
-    id = connection.id()
+    local conn_id = connection.id()
+
     -- wakeup all waiters
     while true do
-        waiter = box.space._queue_consumers.index.pk:min{id}
+        local waiter = box.space._queue_consumers.index.pk:min{conn_id}
```
Totktonada, line 414: Just to understand the idea: consumers are still identified by a connection id rather than queue session id? Can you explain, why it is different from identifying a task ownership?

README.md
```diff
@@ -372,6 +373,20 @@ Available `options`:
 * `ttr` - time to release in seconds. The time after which, if there is no active
 connection in the session, it will be released with all its tasks.
 
+## Session identify
+
+```lua
+queue.identify(session_uuid)
+```
+In the queue the session has a unique UUID and one session can have many
+connections. Also, the consumer can reconnect to the existing session during the
```
Totktonada, line 382: > one session can have many connections

I would say 'many connections may share one logical session' or kinda. It is a thing of taste, feel free to ignore.

README.md
```diff
@@ -372,6 +373,20 @@ Available `options`:
...
+ `ttr` time.  
+To get the UUID of the current session, call the `queue.identify()`
+without parameters.  
+To connect to the existing session, call the `queue.identify(session_uuid)`
+with the UUID of the session.  
+In case of failure, an error will be thrown.
```
Totktonada, line 388: Just curious: what kinds of failures may appear here?

Totktonada, line 388: I would add an example for net.box to better describe the idea. I guess we can use [on_connect net.box trigger](https://github.com/tarantool/tarantool/issues/2858) to transparently overcome a reconnect.

README.md
```diff
@@ -372,6 +373,20 @@ Available `options`:
 * `ttr` - time to release in seconds. The time after which, if there is no active
 connection in the session, it will be released with all its tasks.
 
+## Session identify
+
```
Totktonada, line 377: I would highlight the ability we want to provide with this function somewhere at the start of the section, For example:

> Sometimes we need an ability to acknowledge a task after reconnect (because retrying it is undesirable) or even acknowlegde using another connection.

queue/abstract/queue_session.lua
```diff
@@ -0,0 +1,195 @@
...
+    queue_session._on_session_remove = callback
+end
+
+--- Remove a connection from the list of active connections and
+-- release its tasks if necessary.
+local function remove_connection(conn_id)
+    local queue_session_ids = box.space._queue_session_ids
+    local session_uuid = queue_session.get_uuid(conn_id)
+
+    queue_session_ids:delete{conn_id}
```
Totktonada, line 142: Jusr raw idea (maybe for further PRs): it may be helpful to have ability to trace connections and sessions and issue a log entry (on verbose or even debug level) when a connection is added or removed from a session or moved to another session.

README.md
```diff
@@ -372,6 +373,20 @@ Available `options`:
 * `ttr` - time to release in seconds. The time after which, if there is no active
 connection in the session, it will be released with all its tasks.
 
+## Session identify
```
Totktonada, line 376: I'm not sure, but maybe it worth to name the section 'Overcome a reconnect'. That is what a user will actually find.

Totktonada (approved):
I didn't verified the implementation thoroughly, but the idea looks okay for me.

I left some comments: mostly regarding API and presenting the feature in the readme. Some are just to better understand the idea.

queue/abstract.lua
LeonidVas: > Several nits, feel free to ignore.
> 
> 1. I would not mix helpers with API function. I think it is better to move helpers at top.

Accepted.

> 2. Typo: 'Checks is' → 'if'.

"If the task has been taken in a session...."? Are you sure?

> 3. 'Throw an error on failure' — it is unclear from the description, whether taken or not taken task is considered as failure.

Accepted.

> 4. `check_task_on_taken()` sounds strange. Variants: `check_task_is_taken()`, `check_task_is_owned()`, `check_task_ownership()`. Not sure here.

Yep. But `is` typically returns bool.

> 5. Since there is a usage of the function as a predicate, I would split it to `task_is_taken()` and `error_task_is_not_taken()` functions or kinda. Pcalling of a simple helper looks dubios for me. Alternative: `task_is_taken()` / `task_is_owned()`, which returns `is_taken, err`, so it may be placed directly to `assert()`.

I just moved the common logic into a separate function. If we add a wrapper, it changes the current behavior.

queue/abstract.lua
LeonidVas: I disagree.Before `connection.id()` used in one place in a function. Now `conn_id` is used in two places and using local variable let to avoid unnecessary call `connection.id()`.

queue/abstract.lua
LeonidVas: I don't know "why". I write benchmarks and try several variants. AFAIK this is a valid way to choose a more optimal solution.

queue/abstract/driver/fifottl.lua
LeonidVas: OMG. Thank you. i missed it.

queue/util.lua
LeonidVas: MAX_TIMEOUT < TIMEOUT_INFINITY )
I understand you question, but in this task I only moved the common code to a separate place (DRY). It can be changed by a separate PR. I don't see any reasons to change it here.

queue/util.lua
LeonidVas: Accepted.

queue/util.lua
LeonidVas: I thought:"May be this is a nice way to separate methods and attributes."  AFAIU, you don't think so?

queue/util.lua
LeonidVas: Accepted.

queue/abstract/queue_session.lua
LeonidVas: I do not agree. The parent names the child. But maybe you know some pattern that I don't know?
"...unclear moment after the fiber creating" - I would prefer to use `new` here, but you and Oleg don't want to breake compatibility with the old tarantool (<1.10). In this case, the moment is clear.

queue/abstract/queue_session.lua
LeonidVas: If you want to keep the previous settings: you must get them, change and set them after that.
I think this question will be best to discussed by voice.

queue/abstract/queue_session.lua
LeonidVas: Hmm. I will think about it.

queue/abstract/queue_session.lua
LeonidVas: Accepted.

queue/abstract.lua
LeonidVas: Additionally I use `connection.on_disconnect()`

queue/abstract/queue_session.lua
LeonidVas: Not `connection`, `session`.
I think this question will be best to discussed by voice.

queue/abstract/queue_session.lua
LeonidVas: Accepted.

queue/abstract.lua
LeonidVas: Accepted.

queue/abstract.lua
LeonidVas: Several consumers can work in the same session with the same tasks. For example, one consumer can `take` a task and another consumer (in the same session) can `ack` this task.

README.md
LeonidVas: Accepted.

README.md
LeonidVas: For example: Trying to use invalid format UUID, expired UUID.

LeonidVas: As I said earlier, the examples will be added after the feature will be pushed to master.

README.md
LeonidVas: Accepted.

queue/abstract/queue_session.lua
LeonidVas: Sounds good.

PR #140, opened 2020-11-10
fix custom driver registration after reboot

### LINKED ISSUE #137: Fix custom driver registration after box.cfg()
Now if a custom driver is registered after calling `box.cfg()` it causes a problem when the instance will be restarted. The cause of the problem is an attempt to recreate a tube of custom type (which is not registered yet).
It's needed to implement a deferment of creating a custom tube until the custom driver will be registered.

LeonidVas:
Previously, if a custom driver is registered after calling box.cfg() it causes a problem when the instance will be restarted. The cause of the problem is an attempt to recreate a tube of custom type (which is not registered yet).
This bug is now fixed.

Fixes #137

@ChangeLog:  fix custom driver registration after reboot. Previously, if a custom driver is registered after calling box.cfg() it causes a problem when the instance will be restarted. The cause of the problem is an attempt to recreate a tube of custom type (which is not registered yet).

### DIFF
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -570,13 +570,42 @@ function method.start()
     end
 
     for _, tube_tuple in _queue:pairs() do
-        local tube = recreate_tube(tube_tuple)
-        -- gh-66: release all taken tasks on start
-        tube_release_all_tasks(tube)
+        -- Recreate tubes for registered drivers only.
+        -- Check if a driver exists for this type of tube.
+        if queue.driver[tube_tuple[4]] ~= nil then
+            local tube = recreate_tube(tube_tuple)
+            -- gh-66: release all taken tasks on start
+            tube_release_all_tasks(tube)
+        end
     end
 
     session.on_disconnect(queue._on_consumer_disconnect)
     return queue
+end
+
+--- Register the custom driver.
+-- Unlike the "register_driver" method from init.lua, this method
+-- recreates the existing tubes of the registered driver.
+function method.register_driver(driver_name, tube_ctr)
+    if type(tube_ctr.create_space) ~= 'function' or
+        type(tube_ctr.new) ~= 'function' then
+        error('tube control methods must contain functions "create_space"'
+              .. ' and "new"')
+    end
+    if queue.driver[driver_name] then
+        error(('overriding registered driver "%s"'):format(driver_name))
+    end
+    queue.driver[driver_name] = tube_ctr
+
+    -- Recreates the existing tubes of the registered driver.
+    local _queue = box.space._queue
+    for _, tube_tuple in _queue:pairs() do
+        if tube_tuple[4] == driver_name then
+            local tube = recreate_tube(tube_tuple)
+            -- Release all task for tube on start.
+            tube_release_all_tasks(tube)
+        end
+    end
 end
 
 local function build_stats(space)
--- a/queue/init.lua
+++ b/queue/init.lua
@@ -77,6 +77,8 @@ function wrapper_impl(...)
             rawset(queue, name, val)
         end
         abstract.driver = queue.driver
+        -- Now the "register_driver" method from abstract will be used.
+        queue.register_driver = nil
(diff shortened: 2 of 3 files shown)


LeonidVas:
I honestly tried to write a test, but I lost this fight.

LeonidVas:
> I honestly tried to write a test, but I lost this fight.

queue/abstract.lua
```diff
@@ -570,15 +570,43 @@ function method.start()
     end
 
     for _, tube_tuple in _queue:pairs() do
-        local tube = recreate_tube(tube_tuple)
-        -- gh-66: release all taken tasks on start
-        tube_release_all_tasks(tube)
+        -- Recreate tubes for registered drivers only.
+        if queue.driver[tube_tuple[4]] ~= nil then
```
ligurio, line 574: For my taste it would be better to assign `tube_tuple[4]` to a separate variable for clearness.
Although it's  up to you keep it as is or apply a patch.

```
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -571,7 +571,8 @@ function method.start()
 
     for _, tube_tuple in _queue:pairs() do
         -- Recreate tubes for registered drivers only.
-        if queue.driver[tube_tuple[4]] ~= nil then
+        local tube_type = tube_tuple[4]
+        if queue.driver[tube_type] ~= nil then
             local tube = recreate_tube(tube_tuple)
             -- gh-66: release all taken tasks on start
             tube_release_all_tasks(tube)
```

queue/abstract.lua
```diff
@@ -570,15 +570,43 @@ function method.start()
...
     end
 
     session.on_disconnect(queue._on_consumer_disconnect)
     return queue
 end
 
+--- Register the custom driver.
+-- Unlike the "register_driver" method from init.lua, this method
+-- recreates the existing tubes of the registered driver.
+function method.register_driver(driver_name, tube_ctr)
```
ligurio, line 588: I would change function name to something like 'recreate_tubes()' to avoid confusion with `register_driver()` in `init.lua`.

ligurio (changes requested):

queue/abstract.lua
LeonidVas: I understand you pain. I think the best way is to do as in fifottl driver - to use something like enum. I think I will send a similar patch later. But I don't want to create an additional variable for this. As a compromise, I've added a comment.

queue/abstract.lua
LeonidVas: Sorry, but it's impossible. For the client, this is the same method as before calling box.cfg ().

queue/abstract.lua
```diff
@@ -570,15 +570,44 @@ function method.start()
     end
 
     for _, tube_tuple in _queue:pairs() do
-        local tube = recreate_tube(tube_tuple)
-        -- gh-66: release all taken tasks on start
-        tube_release_all_tasks(tube)
+        -- Recreate tubes for registered drivers only.
+        -- Check if a driver exists for this type of tube.
+        if queue.driver[tube_tuple[4]] ~= nil then
```
Totktonada, line 575: How about emitting a warning in the case?

Totktonada (approved):
I didn't looked at the code thoroughly, but the idea looks good for me.

queue/abstract.lua
LeonidVas: I think it will look like garbage when starting the queue instance registering the custom driver after box.cfg (). And will cause unnecessary questions from users. But if you insist, I won't mind.

ligurio (approved):
Leonid, thanks for the patch!

LGTM

queue/abstract.lua
Totktonada: Your point is right (however it depends on how the log entry will be formulated). OTOH, silent no-op may be confusing if a driver registration does not occurs for some reason. I have no strict preference here, so I don't insist.

PR #143, opened 2020-12-17
fix the shadowing of the variable "space"

LeonidVas:
The local variable "space" is defined twice in `recreate_tube()`.
First time it's a name of the space, the second - it's a space object.
If there is no space, we will try to thrown an error with the name of
the space, but `nil` will be used instead.


Totktonada (approved):
Nice, thanks!

Issue #125, opened 2020-07-10
Add an ability to customize task options (and data) when calling tube:release()


rybakit:
[tube:release()](https://github.com/tarantool/queue#releasing-a-task) accepts an optional second argument `opts`, which at this moment can only consist of `delay`. But often when I release a task back into the queue, I need to adjust other task options, such as priority or ttl. Also, in task data I usually store additional options such as retry strategy, retry count, etc, and it would be great if I can adjust them as well when calling `tube:release()`.

LeonidVas:
What do you think about implementing the task "options" update in a separate method like `touch`?

rybakit:
> What do you think about implementing the task "options" update in a separate method like touch?

@LeonidVas Could you elaborate on how you see it, what options will it support? Maybe you have some examples on hand?

LeonidVas:
> > What do you think about implementing the task "options" update in a separate method like touch?
> 
> @LeonidVas Could you elaborate on how you see it, what options will it support? Maybe you have some examples on hand?

Respond to question with counter-questions - not bad)(Вы, таки, не с Одессы будете?)
Frankly, I started thinking about this task after you refer to it in https://github.com/tarantool/queue/pull/149 .
Now I can only share my raw thoughts on this theme:
- I don't know if this is a true way to change input parameters of a `task` (although we already have the `touch` method that can affect `ttl`).
- `release` called by the worker and IMHO it is outside of his competence to change the input parameters of a `task`.
- Now we already have a method which can affect the input parameters of a `task`(`touch`), and maybe it would be logical to expand functionality of this method instead of adding such functionality to the `release` method.

For ascertaining the answers, I should to look around and see how this works in other similar software, and when I asked you a question, I thought you have some knowledge in this area and you would like to share it.

rybakit:
OK, let me explain then why I asked you about giving more details :) When you proposed the new api method, I thought you had already thought about how it would work and what it should accept. For example, will it be able to increment, decrement, or reset tasks `ttr`:

```lua
t:options(task_id, {ttr_inc=60})
```
Will it allow to change the task state:

```lua
t:options(task_id, {state=state.READY})
```
or/and delay:

```lua
t:options(task_id, {delay=42})
```

If so, what do you plan to do with the existing functions that already have this functionality? Deprecate them or keep them?
If not, what options do you actually propose to allow with this new function? Also, what about the custom options that I mentioned in the description? Will it support them and in which form?

[...]

PR #149, opened 2021-04-19
Option to keep the ttl while releasing the task

qdrin:
Adds the option 'touch_ttl' that is assotiated with delay option in the release method of fifottl. Allows keep the task ttl unchanged when it necessary. Might help with retries.

### DIFF
--- a/queue/abstract/driver/fifottl.lua
+++ b/queue/abstract/driver/fifottl.lua
@@ -300,12 +300,18 @@ function method.release(self, id, opts)
     if task == nil then
         return
     end
+    if opts.touch_ttl == nil then opts.touch_ttl = true end
     if opts.delay ~= nil and opts.delay > 0 then
-        task = self.space:update(id, {
+        local upd = {
             { '=', i_status, state.DELAYED },
-            { '=', i_next_event, util.event_time(opts.delay) },
+            { '=', i_next_event, util.event_time(opts.delay) }
+        }
+        if opts.touch_ttl then
+            table.insert(upd,
             { '+', i_ttl, util.time(opts.delay) }
-        })
+            )
+        end
+        task = self.space:update(id, upd)
     else
         task = self.space:update(id, {
             { '=', i_status, state.READY },
--- a/queue/abstract/driver/fifottl.lua
+++ b/queue/abstract/driver/fifottl.lua
@@ -300,13 +300,12 @@ function method.release(self, id, opts)
     if task == nil then
         return
     end
-    if opts.touch_ttl == nil then opts.touch_ttl = true end
     if opts.delay ~= nil and opts.delay > 0 then
         local upd = {
             { '=', i_status, state.DELAYED },
             { '=', i_next_event, util.event_time(opts.delay) }
         }
-        if opts.touch_ttl then
+        if not opts.keep_ttl then
             table.insert(upd,
             { '+', i_ttl, util.time(opts.delay) }
             )
--- a/README.md
+++ b/README.md
@@ -551,6 +551,12 @@ Example: queue.tube.list_of_sites:release(15, {delay=10})
 Note: in the above example, the delay option means
 "the task cannot be executed again for 10 seconds".
 
+Example: queue.tube.list_of_sites:release(15, {delay=10, touch_ttl=false})
+
+Note: in the above example, the delay option means
+"the task cannot be executed again for 10 seconds" and
+the associated touch_ttl option means "the task ttl stays unchanged"
+
 ## Peeking at a task
 
 ```lua
--- a/README.md
+++ b/README.md
@@ -551,11 +551,11 @@ Example: queue.tube.list_of_sites:release(15, {delay=10})
 Note: in the above example, the delay option means
 "the task cannot be executed again for 10 seconds".
 
-Example: queue.tube.list_of_sites:release(15, {delay=10, touch_ttl=false})
+Example: queue.tube.list_of_sites:release(15, {delay=10, keep_ttl=true})
 
 Note: in the above example, the delay option means
 "the task cannot be executed again for 10 seconds" and
-the associated touch_ttl option means "the task ttl stays unchanged"
+the associated keep_ttl option means "the task ttl stays unchanged"
 
 ## Peeking at a task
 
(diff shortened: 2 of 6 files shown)


queue/abstract/driver/fifottl.lua
```diff
@@ -300,12 +300,18 @@ function method.release(self, id, opts)
...
-            { '=', i_next_event, util.event_time(opts.delay) },
+            { '=', i_next_event, util.event_time(opts.delay) }
+        }
+        if opts.touch_ttl then
+            table.insert(upd,
             { '+', i_ttl, util.time(opts.delay) }
-        })
+            )
+        end
+        task = self.space:update(id, upd)
```
Totktonada, line 314: I would name a new option in a way to make its default value `false`: say, `keep_ttl`. It looks more natural for me.

I guess the code hunk may be written in a simpler way (didn't tested):

```lua
task = self.space:update(id, {
    { '=', i_status, state.DELAYED },
    { '=', i_next_event, util.event_time(opts.delay) },
    not opts.keep_ttl and { '+', i_ttl, util.time(opts.delay) } or nil,
})
```

queue/abstract/driver/fifottl.lua
```diff
@@ -300,12 +300,18 @@ function method.release(self, id, opts)
     if task == nil then
         return
     end
+    if opts.touch_ttl == nil then opts.touch_ttl = true end
```
Totktonada, line 303: It would be good to support it in 'utubettl' driver too.

Totktonada, line 303: We usually cover all new functionality with a test. Aren't you mind to add one? They're in the `t/` directory.

qdrin: Covered in new commit. Sorry for that

qdrin: Added in new commit. Thanks for Your idea

queue/abstract/driver/fifottl.lua
qdrin: Thanks for the fine advise. Renamed

rybakit:
IMHO, introducing a new option that solely depends on another option is really not the best solution to the problem. I'm afraid it will be a source of confusion for users, because the API allows the "associated" option to be passed alone and, when allowed, is likely to be misused. So I expect to see in the future something like

```
t:release(15, {keep_ttl=true})
```
which makes no sense. Moreover, when https://github.com/tarantool/queue/issues/125 gets implemented, we might get

```
t:release(15, {delay=10, keep_ttl=true, ttl=10})
```
which is a valid combination of options, although it looks very weird and confusing.

As an alternative, I would consider adding a new standalone option, something like `delay_with_ttl` or `delay_wo_ttl` (or any other name that works best), so at least it can be used alone. Also, we might think of forbidding passing both `delay` and `delay_with_ttl` or at least clearly define the precedence of these options.

Another alternative is to introduce a new api method.

rybakit:
After thinking more about the issue, I believe it can be solved in the scope of #125. If we had the ability to pass more options to `release()` we could do

```
t:release(15, {delay=10, ttl=nil})
```

where `ttl=nil` indicates that we don't want to update ttl.

qdrin:
> IMHO, introducing a new option that solely depends on another option is really not the best solution to the problem. I'm afraid it will be a source of confusion for users, because the API allows the "associated" option to be passed alone and, when allowed, is likely to be misused. So I expect to see in the future something like
> 
> ```
> t:release(15, {keep_ttl=true})
> ```
> 
> which makes no sense (however, it gives a false impression that it somehow affects the task ttl). Moreover, when #125 gets implemented, we might get
> 
> ```
> t:release(15, {delay=10, keep_ttl=true, ttl=10})
> ```
> 
> which is a valid combination of options, although it looks very weird and confusing.
> 
> As an alternative, I would consider adding a new standalone option, something like `delay_with_ttl` or `delay_wo_ttl` (or any other name that works best), so at least it can be used alone. Also, we might think of forbidding passing both `delay` and `delay_with_ttl` or at least clearly define the precedence of these options.
> 
> Another alternative is to introduce a new api method.

Perhaps the option is not the best solution. But in the current implementation it doesn't seem confusive as the only case with ttl update is the release.
I agree the #125 is fine and covers much more cases. So it's implementation seems to be more complicated and i suppose deserves the particular driver. Meenwhile my purpose was to give some new uses to current one with minimum efforts.

LeonidVas:
HI! Thank you for the patch.
I have several questions:
- Why can't you just use the [`touch`](https://github.com/tarantool/queue#increasing-ttr-andor-ttl-for-tasks) method for this purpose?
- Could you to describe few use case of the feature?

qdrin:
> IMHO, introducing a new option that solely depends on another option is really not the best solution to the problem. I'm afraid it will be a source of confusion for users, because the API allows the "associated" option to be passed alone and, when allowed, is likely to be misused. So I expect to see in the future something like
> 
> ```
> t:release(15, {keep_ttl=true})
> ```
> 
> which makes no sense (however, it gives a false impression that it somehow affects the task ttl). Moreover, when #125 gets implemented, we might get
> 
> ```
> t:release(15, {delay=10, keep_ttl=true, ttl=10})
> ```
> 
> which is a valid combination of options, although it looks very weird and confusing.
> 
> As an alternative, I would consider adding a new standalone option, something like `delay_with_ttl` or `delay_wo_ttl` (or any other name that works best), so at least it can be used alone. Also, we might think of forbidding passing both `delay` and `delay_with_ttl` or at least clearly define the precedence of these options.
> 
> Another alternative is to introduce a new api method.

Perhaps the option is not the best solution. But in the current implementation it doesn't seem confusive as the only case with ttl update is the release.
I agree the #125 is fine and covers much more cases. So it's implementation seems to be more complicated and i suppose deserves the particular driver. Meenwhile my purpose was to give some new uses to current one with minimum efforts.


>     * Could you to describe few use case of the feature?

qdrin:
Sorry it seems i close the request accidently. Let's me reopen

LeonidVas:
> My case is to retry failed tasks (http requests) every 30 seconds for 6-minutes period. So I need ttl to be constant. Shure if touch() could decrease ttl it would solve the issue along with "classic" release()

Is it close to `ttr = 30` with `ttl = 360`. Did I understand your case correctly?

LeonidVas:
Well, thanks for the PR, but it looks like we haven't decided yet whether such functionality is needed in the module and how exactly the API of this functionality should look like. I think it would be good to resolve these questions in discussion #125, and then start implementing the feature.

PR #159, opened 2021-12-17
fix fifottl bury

GoWebProd:
Now the task will be removed from the queue if a TTR has been installed for it. Fix allows you to delete a task only if it has a TTL


LeonidVas (approved):
Hi! Thank you for the patch.
I updated it, added a comment, test, commit message and asked colleagues to review the patch.

0x501D (approved):

PR #164, opened 2022-03-03
add local queue

GoWebProd:
Ability to create not temporary queues but without replication


LeonidVas (commented):
Hi! Thank you for the patch.
Sounds interesting, but I think to wait until the work on the https://github.com/tarantool/queue/pull/165 is finished to see how best to join it, there may be some nuances.

LeonidVas:
We discussed with the team whether this functionality could be useful, and could not present some good scenarios where it could be used.
So I decided to close the PR. Fill free to reopen the PR if you can suggest a couple.

PR #186, opened 2022-08-29
abstract: fix switch in_replicaset value

### LINKED ISSUE #185: Unable to queue.cfg{in_replicaset = true} with Tarantool 1.10
How-to reproduce:
```bash
$ tarantool config.lua 
...
2022-08-25 18:00:53.885 [78123] main/101/config.lua I> Started queue state fiber
2022-08-25 18:00:53.887 [78123] main txn.c:438 E> ER_UNSUPPORTED: DDL does not support multi-statement transactions
2022-08-25 18:00:53.887 [78123] main F> fatal error, exiting the event loop
```
Reproducer:
```bash
#!/usr/bin/env tarantool

box.cfg{}
queue = require("queue")
queue.cfg{in_replicaset = true}
```

Caused by:

https://github.com/tarantool/queue/blob/fd3616c9ba121bbaf858d79c0e1a21fa3d3727e4/queue/abstract.lua#L666-L667

https://github.com/tarantool/queue/blob/fd3616c9ba121bbaf858d79c0e1a21fa3d3727e4/queue/abstract/queue_session.lua#L65-L66

oleg-jukovec:
Execution of non-yielding DDL statements in transactions has been allowed only since Tarantool 2.2.1 [1]. We should avoid using DDL statements in transactions for older versions.

1. https://github.com/tarantool/tarantool/commit/f266559b167b4643c377724eebde9651877d6cc9

Closes #185 

This patchset includes a few fixes, mostly typos. So I recommend to review commits one by one. I can extract it into a separate pull request, if you wish, but the fixes are really small.

### DIFF
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -11,6 +11,10 @@ local qc       = require('queue.compat')
 local num_type = qc.num_type
 local str_type = qc.str_type
 
+-- since:
+-- https://github.com/tarantool/tarantool/commit/f266559b167b4643c377724eebde9651877d6cc9
+local ddl_txn_supported = qc.check_version({2, 2, 1})
+
 -- The term "queue session" has been added to the queue. One "queue session"
 -- can include many connections (box.session). For clarity, the box.session
 -- will be referred to as connection below.
@@ -668,12 +672,22 @@ local function switch_in_replicaset(replicaset_mode)
         box.space._queue_taken_2_mgr:insert(tuple)
     end
 
-    box.space._queue_taken_2:drop()
-    box.space._queue_taken_2_mgr:rename('_queue_taken_2')
+    if ddl_txn_supported then
+		-- We can do it inside a transaction.
+        box.space._queue_taken_2:drop()
+        box.space._queue_taken_2_mgr:rename('_queue_taken_2')
+    end
 
     local status, err = pcall(box.commit)
     if not status then
         error(('Error migrate _queue_taken_2: %s'):format(tostring(err)))
+    end
+
+    if not ddl_txn_supported then
+		-- Do it outside a transaction becase DDL does not support
+		-- multi-statement transactions.
+        box.space._queue_taken_2:drop()
+        box.space._queue_taken_2_mgr:rename('_queue_taken_2')
     end
 end
 
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -579,8 +579,7 @@ function method.create_tube(tube_name, tube_type, opts)
 
     local replicaset_mode = queue.cfg['in_replicaset'] or false
     if replicaset_mode and opts.temporary then
-        log.error("Cannot create temporary tube in replicaset mode")
-        return
+        error("Cannot create temporary tube in replicaset mode")
     end
 
     local driver = queue.driver[tube_type]
--- a/queue/abstract/queue_session.lua
+++ b/queue/abstract/queue_session.lua
@@ -7,6 +7,9 @@ local qc       = require('queue.compat')
 local num_type = qc.num_type
 local str_type = qc.str_type
 
+-- since:
+-- https://github.com/tarantool/tarantool/commit/f266559b167b4643c377724eebde9651877d6cc9
+local ddl_txn_supported = qc.check_version({2, 2, 1})
 
 local queue_session = {}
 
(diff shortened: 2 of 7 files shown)


0x501D (approved):
LGTM

queue/abstract.lua
```diff
@@ -11,6 +11,10 @@ local qc       = require('queue.compat')
 local num_type = qc.num_type
 local str_type = qc.str_type
 
+-- since:
+-- https://github.com/tarantool/tarantool/commit/f266559b167b4643c377724eebde9651877d6cc9
+local is_ddl_txn_supported = qc.check_version({2, 2, 1})
```
LeonidVas, line 16: I suggest not using the prefix `is_` as we do not use it in boolean `temporary`, `unique`...

queue/abstract.lua
```diff
@@ -668,13 +672,20 @@ local function switch_in_replicaset(replicaset_mode)
         box.space._queue_taken_2_mgr:insert(tuple)
     end
 
-    box.space._queue_taken_2:drop()
-    box.space._queue_taken_2_mgr:rename('_queue_taken_2')
+    if is_ddl_txn_supported then
```
LeonidVas, line 675: Please add a comment here. Because for a new developer it will look like:
```
if x then
    do some action
end
...

if not x then
    perform the same action that would be done if x were true
end
```

queue/abstract/queue_session.lua
```diff
@@ -7,6 +7,9 @@ local qc       = require('queue.compat')
 local num_type = qc.num_type
 local str_type = qc.str_type
 
+-- since:
+-- https://github.com/tarantool/tarantool/commit/f266559b167b4643c377724eebde9651877d6cc9
+local is_ddl_txn_supported = qc.check_version({2, 2, 1})
```
LeonidVas, line 12: I suggest not using the prefix `is_` as we do not use it in boolean `temporary`, `unique`...

t/210-cfg-in-replicaset.t
```diff
@@ -0,0 +1,43 @@
...
+    test:ok(status, 'in_replicaset = true switched')
+    local status, _ = pcall(queue.cfg, {in_replicaset = false})
+    test:ok(status, 'in_replicaset = false switched')
+    local status, _ = pcall(queue.cfg, {in_replicaset = true})
+    test:ok(status, 'in_replicaset = true switched')
+    local status, _ = pcall(queue.cfg, {in_replicaset = false})
+    test:ok(status, 'in_replicaset = false switched')
+end)
+
+test:test('Check error create temporary tube', function(test)
```
LeonidVas, line 26: Why did you decide to move this test?

LeonidVas (changes requested):
Hi! Thank you for the patchset.
Good catch.
About commit messages:
* "abstract: error if failed to create a temp tube" -> "abstract: throw an error if failed to create a temp tube"
* "The create_table method returns" -> "The create_tube method returns"
* "fix: switch in_replicaset value with tnt < 2.2.1" - "fix" not is a subsystem.

queue/abstract.lua
oleg-jukovec: Thank you, as you wish.

queue/abstract.lua
oleg-jukovec: Added/

queue/abstract/queue_session.lua
oleg-jukovec: Thank you, as you wish. `is_ddl_txn_supported` -> `ddl_txn_supported`

t/210-cfg-in-replicaset.t
oleg-jukovec: It looks like a test for a local instance. It has no external dependencies.

oleg-jukovec:
> About commit messages:
>     * "abstract: error if failed to create a temp tube" -> "abstract: throw an error if failed to create a temp tube"
>     * "The create_table method returns" -> "The create_tube method returns"
>     * "fix: switch in_replicaset value with tnt < 2.2.1" - "fix" not is a subsystem.

Thank you for your comment. I tried to do the best:

https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message

LeonidVas (approved):
LGTM

PR #188, opened 2022-09-06
test: fix flaky 050-ttl.t

### LINKED ISSUE #187: Flaky t/050-ttl.t
83fe-391c-1e78-88a6

https://github.com/tarantool/tarantool/runs/8191312280
https://github.com/tarantool/queue/runs/8204532401

oleg-jukovec:
We use system clock in the internal implementation for TTL, not
monotonic clock. The test may fail when changing the system clock. In
addition, fiber.sleep() may wake up earlier than planned by the
system clock.

We should to handle these situations in the test.

Closes #187


t/050-ttl.t
```diff
@@ -17,13 +17,26 @@ local ttl = 0.1
 
 test:ok(queue, 'queue is loaded')
 
+local function test_take_after_ttl(test, tube, ttl)
```
LeonidVas, line 20: - Can't we just check the result after `ttl + ttl/2`
- Use of skip is invalid if `after - before < ttl * 1000000` we must still wait and recheck the result

LeonidVas (changes requested):
Hi, thank you for the patch.
`The test may fail when changing the system clock` - I think we can neglect this probability in the test.

t/050-ttl.t
oleg-jukovec: > I think we can neglect this probability in the test.

I think it wil be better to give a meaningful error.

> * Can't we just check the result after `ttl + ttl/2`

We can, but it depends on what we decide to do with system clock change.

>     * Use of skip is invalid if `after - before < ttl * 1000000` we must still wait and recheck the result

Please see the update. If you don't like it, then I'll just replace `fibe.sleep(ttl)` with `fiber.sleep(ttl + ttl/2)`. It will almost always work and will be easier.

0x501D (approved):
LGTM

LeonidVas (approved):

PR #189, opened 2022-09-08
cd: downgrade a runner image to ubuntu-18.04

oleg-jukovec:
Due to a bug [1], packaging for Centos 7 does not work on ubuntu-latest. We need to downgrade a GitHub Actions runner image version until the problem is fixed.

1. https://github.com/packpack/packpack/issues/145

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -9,7 +9,9 @@ jobs:
     if: |
       github.event_name == 'push' ||
       github.event.pull_request.head.repo.full_name != github.repository
-    runs-on: ubuntu-latest
+    # Packaging for CentOS 7 does not work with other versions, see:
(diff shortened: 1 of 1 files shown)


.github/workflows/packaging.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     if: |
       github.event_name == 'push' ||
       github.event.pull_request.head.repo.full_name != github.repository
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
```
Totktonada, line 12: I would leave a comment that briefly describes the problem and links the issue. Otherwise it is easy to conclude that this choice is arbitrary and step into the problem again.

LeonidVas (approved):

PR #196, opened 2022-11-03
Fix packaging for CentOS7

LeonidVas:
Fix the problem with sudoers for CentOS7.

For more details see:
https://github.com/packpack/packpack/issues/145

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -9,9 +9,9 @@ jobs:
     if: |
       github.event_name == 'push' ||
       github.event.pull_request.head.repo.full_name != github.repository
-    # Packaging for CentOS 7 does not work with other versions, see:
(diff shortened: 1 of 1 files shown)


.github/workflows/packaging.yml
```diff
@@ -61,6 +61,28 @@ jobs:
         # that all tags will exists always.
         run: git fetch --tags -f
 
+      # We use a certain version of Docker because otherwise we have a problem
```
oleg-jukovec, line 64:
```suggestion
      # We use a particular version of Docker because otherwise we have a problem
```
or specific?

oleg-jukovec (approved):
Thank you for the fix!

.github/workflows/packaging.yml
LeonidVas: I choose the "specific".

PR #198, opened 2022-11-29
Remove printing secrets to the log

LeonidVas:
According to [1] printing secrets to the log is not a best practice.

[1] https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -77,10 +77,18 @@ jobs:
             -u ${RWS_AUTH} \
             -F product=${PRODUCT_NAME}"
 
+          # We don't want to try to print secrets to the log, but we want
+          # to print a "curl" command to see what's going on.
+          CURL_CMD_ECHO="curl -LfsS \
+            -X PUT ${RWS_URL_PART}/${OS}/${DIST} \
+            -u *** \
+            -F product=${PRODUCT_NAME}"
+
           for f in $(ls -I '*build*' -I '*.changes' ./build); do
             CURL_CMD+=" -F $(basename ${f})=@./build/${f}"
+            CURL_CMD_ECHO+=" -F $(basename ${f})=@./build/${f}"
           done
 
-          echo ${CURL_CMD}
+          echo ${CURL_CMD_ECHO}
 
           ${CURL_CMD}


.github/workflows/packaging.yml
```diff
@@ -81,6 +81,6 @@ jobs:
             CURL_CMD+=" -F $(basename ${f})=@./build/${f}"
           done
 
-          echo ${CURL_CMD}
+          echo "Upload packages..."
```
Totktonada, line 84: I just checked: the secret is not actually shown in the logs:

> curl -LfsS -X PUT https://rws.tarantool.org/tarantool-modules/debian/stretch -u *** -F product=tarantool-queue -F tarantool-queue-1.2.2.0.tar.xz=@./build/tarantool-queue-1.2.2.0.tar.xz -F tarantool-queue_1.2.2.0-1.debian.tar.xz=@./build/tarantool-queue_1.2.2.0-1.debian.tar.xz -F tarantool-queue_1.2.2.0-1.dsc=@./build/tarantool-queue_1.2.2.0-1.dsc -F tarantool-queue_1.2.2.0-1_all.deb=@./build/tarantool-queue_1.2.2.0-1_all.deb -F tarantool-queue_1.2.2.0.orig.tar.xz=@./build/tarantool-queue_1.2.2.0.orig.tar.xz

The commit title/message may be confusing.

Totktonada, line 84: I would keep echoing the command, but without actual username/password. It may be useful for debugging.

How about something like the following?

```bash
CURL_CMD=<...>
CURL_CMD_ECHO=<...>

for <...>; do
    CURL_CMD=<...>
    CURL_CMD_ECHO=<...>
done

echo ${CURL_CMD_ECHO}
${CURL_CMD}
```

Sooner or later this come should go into rws'es GitHub action, so some bloating is not critical.

LeonidVas: The commit message has been updated.

.github/workflows/packaging.yml
```diff
@@ -77,10 +77,17 @@ jobs:
             -u ${RWS_AUTH} \
             -F product=${PRODUCT_NAME}"
 
+        # We don't want to try to print secrets to the log, but we want
+        # to print a "curl" command to see what's going on.
+          CURL_CMD_ECHO="curl -LfsS \
+            -X PUT ${RWS_URL_PART}/${OS}/${DIST} \
+            -F product=${PRODUCT_NAME}"
```
Totktonada, line 84: I would add `-u ***` or `-u <...>` or `-u ${RWS_AUTH}` explicitly. The latter also allows to define the variable and copy-paste the command.

LeonidVas: I choose `-u ***`

Totktonada (approved):

PR #203, opened 2023-03-06
replicaset_mode: fix queue.tube indexes

### LINKED ISSUE #202: Failure when trying to use the tube after switching back to the master
When using the queue in the replica set, when switching from master to replica and then back again, the module does not allow to work with tubes created on the replica.

The initial settings are taken from [here](https://github.com/tarantool/queue#queue-and-replication).

master:

`box.cfg({read_only = true})`

replica:

```
box.cfg({read_only = false})
queue.create_tube('tube', 'fifo')
box.cfg({read_only = true})

```
master:

`box.cfg({read_only = false})`

and now we trying to use the `tube'.

```
tarantool> queue.tube.tube:put("a1")
---
- error: '[string "return queue.tube.tube:put("a1")"]:1: attempt to index field ''tube''
    (a nil value)'
...
```

0x501D:
In replicaset mode, tubes can be created and deleted on different nodes. Accordingly, it is necessary to rebuild the queue.tube index.

Close #202

### DIFF
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -520,10 +520,39 @@ function method._on_consumer_disconnect()
     session.disconnect(conn_id)
 end
 
+-- function takes tuples and recreates tube
+local function recreate_tube(tube_tuple)
+    local name, id, space_name, tube_type, opts = tube_tuple:unpack()
+
+    local driver = queue.driver[tube_type]
+    if driver == nil then
+        error("Unknown tube type " .. tostring(tube_type))
+    end
+
+    local space = box.space[space_name]
+    if space == nil then
+        error(("Space '%s' doesn't exists"):format(space_name))
+    end
+    return make_self(driver, space, name, tube_type, id, opts)
+end
+
 -- Function takes new queue state.
 -- The "RUNNING" and "WAITING" states do not require additional actions.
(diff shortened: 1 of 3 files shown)


LeonidVas (approved):
Hi! Thank you for the patch.
`Close # ...` -> `Closes #`
LGTM

0x501D:
> Hi! Thank you for the patch. `Close # ...` -> `Closes #` LGTM

Fixed.

queue/abstract.lua
```diff
@@ -520,10 +520,39 @@ function method._on_consumer_disconnect()
...
+        -- Accordingly, it is necessary to rebuild the queue.tube index.
+        if replicaset_mode then
+            for _, tube_name in pairs(queue.tube()) do
+                queue.tube[tube_name] = nil
+            end
+            for _, tube_tuple in box.space._queue:pairs() do
+                if queue.driver[tube_tuple[4]] ~= nil then
+                    recreate_tube(tube_tuple)
+                end
+            end
```
oleg-jukovec, line 554: You could extract it in a separate function: `recreate_tubes` - the code appears at least two times in the file.

oleg-jukovec (approved):
Thank you for the patch!

PR #209, opened 2023-06-14
driver: fix yield in fifottl/utubettl

### LINKED ISSUE #208: Check for infinite loop and sleep after 1000 success iterations for fifottl
It seems that queue/fifottl driver have the same problem (at least I don't see the `processed` counter reset):

https://github.com/tarantool/sharded-queue/pull/49

We need to check the code and fix it if necessary.

oleg-jukovec:
After 1000 processed tuples an iteration fiber yields after each processed tuple. The problem was fixed by resetting a counter of processed tuples without yield.

Closes #208

### DIFF
--- a/queue/abstract/driver/fifottl.lua
+++ b/queue/abstract/driver/fifottl.lua
@@ -144,6 +144,7 @@ local function fifottl_fiber_iteration(self, processed)
     if estimated > 0 or processed > 1000 then
         -- free refcounter
         estimated = estimated > 0 and estimated or 0
+        processed = 0
         self.cond:wait(estimated)
     end
 
--- a/queue/abstract/driver/utubettl.lua
+++ b/queue/abstract/driver/utubettl.lua
@@ -152,6 +152,7 @@ local function utubettl_fiber_iteration(self, processed)
         -- free refcounter
         estimated = processed > 1000 and 0 or estimated
         estimated = estimated > 0 and estimated or 0
+        processed = 0
         self.cond:wait(estimated)
     end
 


0x501D (approved):

queue/abstract/driver/fifottl.lua
```diff
@@ -144,6 +144,7 @@ local function fifottl_fiber_iteration(self, processed)
     if estimated > 0 or processed > 1000 then
         -- free refcounter
         estimated = estimated > 0 and estimated or 0
+        processed = 0
```
LeonidVas, line 147: Hmm, I'm not sure this is a good place to reset the counter, because the function returns this value. Wouldn't it be better to reset the counter from the outside?

LeonidVas (changes requested):
Hi! thank you for the patch.
About the title of the commit: "bugfix" not is some subsystem, replace it with something like: "driver: fix yield in fifottl/utubettl".

queue/abstract/driver/fifottl.lua
oleg-jukovec: I don't think so:

Now only one function uses the `processed` variable. In fact the return value is just a state save. If we decide to reset the counter in an external function, then we will spread the responsibility for the variable (and yield + reset) between two functions.

But we could extract wait + reset code into a separate method.

oleg-jukovec:
> with something like: "driver: fix yield in fifottl/utubettl".

Updated.

LeonidVas (approved):

PR #211, opened 2023-07-06
driver: fix duplicate id error with mvvc on

### LINKED ISSUE #207: IDs of task can duplicated when mvсс is on

better0fdead:
Taking the maximum of the index is an implicit transactions, so it is
always done with 'read-confirmed' mvcc isolation level.
It can lead to errors when trying to make parallel 'put' calls with mvcc enabled.
It is hapenning because 'max' for several puts in parallel will be the same since
read confirmed isolation level makes visible all transactions that finished the commit.
To fix it we wrap it with box.begin/commit and set right isolation level.
Current fix does not resolve that bug in situations when we already are in transaction
since it will open nested transactions.

Further work will be in another PR.
Part of #207

### DIFF
--- a/queue/abstract/driver/fifo.lua
+++ b/queue/abstract/driver/fifo.lua
@@ -68,7 +68,27 @@ end
 
 -- put task in space
 function method.put(self, data, opts)
-    local max = self.space.index.task_id:max()
+    local max
+
+    -- Taking the maximum of the index is an implicit transactions, so it is
+    -- always done with 'read-confirmed' mvcc isolation level.
+    -- It can lead to errors when trying to make parallel 'put' calls with mvcc enabled.
+    -- It is hapenning because 'max' for several puts in parallel will be the same since
+    -- read confirmed isolation level makes visible all transactions that finished the commit.
+    -- To fix it we wrap it with box.begin/commit and set right isolation level.
+    -- Current fix does not resolve that bug in situations when we already are in transaction
+    -- since it will open nested transactions.
+    -- See https://github.com/tarantool/queue/issues/207
+    -- See https://www.tarantool.io/ru/doc/latest/concepts/atomic/txn_mode_mvcc/
+
+    if box.cfg.memtx_use_mvcc_engine and (not box.is_in_txn()) then
+        box.begin({txn_isolation = 'read-committed'})
+        max = self.space.index.task_id:max()
+        box.commit()
+    else
+        max = self.space.index.task_id:max()
+    end
+
     local id = max and max[1] + 1 or 0
     local task = self.space:insert{id, state.READY, data}
     self.on_task_change(task, 'put')
@@ -77,7 +97,25 @@ end
 
 -- take task
 function method.take(self)
-    local task = self.space.index.status:min{state.READY}
+    local task
+    -- Taking the minimum is an implicit transactions, so it is
+    -- always done with 'read-confirmed' mvcc isolation level.
+    -- It can lead to errors when trying to make parallel 'take' calls with mvcc enabled.
+    -- It is hapenning because 'min' for several takes in parallel will be the same since
+    -- read confirmed isolation level makes visible all transactions that finished the commit.
+    -- To fix it we wrap it with box.begin/commit and set right isolation level.
+    -- Current fix does not resolve that bug in situations when we already are in transaction
+    -- since it will open nested transactions.
+    -- See https://github.com/tarantool/queue/issues/207
+    -- See https://www.tarantool.io/ru/doc/latest/concepts/atomic/txn_mode_mvcc/
+    if box.cfg.memtx_use_mvcc_engine and (not box.is_in_txn()) then
+        box.begin({txn_isolation = 'read-committed'})
+        task = self.space.index.status:min{state.READY}
+        box.commit()
+    else
+        task = self.space.index.status:min{state.READY}
+    end
+
     if task ~= nil and task[2] == state.READY then
         task = self.space:update(task[1], { { '=', 2, state.TAKEN } })
         self.on_task_change(task, 'take')
--- a/queue/abstract/driver/utubettl.lua
+++ b/queue/abstract/driver/utubettl.lua
@@ -217,7 +217,27 @@ end
 
 -- put task in space
 function method.put(self, data, opts)
-    local max = self.space.index.task_id:max()
+    local max
+
+    -- Taking the maximum of the index is an implicit transactions, so it is
+    -- always done with 'read-confirmed' mvcc isolation level.
+    -- It can lead to errors when trying to make parallel 'put' calls with mvcc enabled.
+    -- It is hapenning because 'max' for several puts in parallel will be the same since
+    -- read confirmed isolation level makes visible all transactions that finished the commit.
+    -- To fix it we wrap it with box.begin/commit and set right isolation level.
+    -- Current fix does not resolve that bug in situations when we already are in transaction
+    -- since it will open nested transactions.
+    -- See https://github.com/tarantool/queue/issues/207
+    -- See https://www.tarantool.io/ru/doc/latest/concepts/atomic/txn_mode_mvcc/
+
+    if box.cfg.memtx_use_mvcc_engine and (not box.is_in_txn()) then
+        box.begin({txn_isolation = 'read-committed'})
+        max = self.space.index.task_id:max()
+        box.commit()
+    else
+        max = self.space.index.task_id:max()
+    end
+
     local id = max and max[i_id] + 1 or 0
 
     local status
@@ -280,7 +300,25 @@ function method.take(self)
             break
         elseif not is_expired(t) then
             local next_event = util.time() + t[i_ttr]
-            local taken = self.space.index.utube:min{state.TAKEN, t[i_utube]}
+            local taken
+            -- Taking the minimum is an implicit transactions, so it is
+            -- always done with 'read-confirmed' mvcc isolation level.
+            -- It can lead to errors when trying to make parallel 'take' calls with mvcc enabled.
+            -- It is hapenning because 'min' for several takes in parallel will be the same since
+            -- read confirmed isolation level makes visible all transactions that finished the commit.
+            -- To fix it we wrap it with box.begin/commit and set right isolation level.
+            -- Current fix does not resolve that bug in situations when we already are in transaction
+            -- since it will open nested transactions.
+            -- See https://github.com/tarantool/queue/issues/207
+            -- See https://www.tarantool.io/ru/doc/latest/concepts/atomic/txn_mode_mvcc/
+            if box.cfg.memtx_use_mvcc_engine and (not box.is_in_txn()) then
+                box.begin({txn_isolation = 'read-committed'})
+                taken = self.space.index.utube:min{state.TAKEN, t[i_utube]}
+                box.commit()
+            else
+                taken = self.space.index.utube:min{state.TAKEN, t[i_utube]}
+            end
+
             if taken == nil or taken[i_status] ~= state.TAKEN then
                 t = self.space:update(t[1], {
                     { '=', i_status, state.TAKEN },
--- a/t/220-mvcc.t
+++ b/t/220-mvcc.t
@@ -1,0 +1,69 @@
+#!/usr/bin/env tarantool
+
+local qc = require('queue.compat')
+local log   = require('log')
+if not qc.check_version({2, 6, 1}) then
+    log.info('Tests skipped, tarantool version < 2.6.1')
+    return
+end
+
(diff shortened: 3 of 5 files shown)


queue/abstract/driver/utubettl.lua
```diff
@@ -215,9 +215,36 @@ function method.normalize_task(self, task)
...
+    -- taking the maximum of the index is an implicit transactions, so it is
+    -- always done with 'read-confirmed' mvcc isolation level.
+    -- It can lead to errors when trying to make parallel 'put' calls with mvcc enabled.
+    -- It is hapenning because 'max' for several puts in parallel will be the same since
+    -- read confirmed isolation level makes visible all transactions that finished the commit.
+    -- To fix it we wrap it with box.begin/commit and set right isolation level.
+    -- See https://github.com/tarantool/queue/issues/207
+    -- See https://www.tarantool.io/ru/doc/latest/concepts/atomic/txn_mode_mvcc/
+    if check_mvcc_state() then
+        box.begin({txn_isolation = 'read-committed'})
```
olegrok, line 241: Is it possible that external transaction will be opened by user? In that case you will open nested transaction. But Tarantool doesn't support it.

queue/abstract/driver/fifo.lua
```diff
@@ -66,9 +66,36 @@ function method.normalize_task(self, task)
...
 
+-- check if mvcc is enabled
+local function check_mvcc_state()
+    local cfg = box.cfg
+    if cfg['memtx_use_mvcc_engine'] == nil  or cfg['memtx_use_mvcc_engine'] == false then
+        return false
+    end
+
+    return true
+end
```
Totktonada, line 77: Just `if box.cfg.memtx_use_mvcc_engine then <...> end` is equivalent and simpler.

queue/abstract/driver/fifo.lua
```diff
@@ -66,9 +66,36 @@ function method.normalize_task(self, task)
...
+    -- It can lead to errors when trying to make parallel 'put' calls with mvcc enabled.
+    -- It is hapenning because 'max' for several puts in parallel will be the same since
+    -- read confirmed isolation level makes visible all transactions that finished the commit.
+    -- To fix it we wrap it with box.begin/commit and set right isolation level.
+    -- See https://github.com/tarantool/queue/issues/207
+    -- See https://www.tarantool.io/ru/doc/latest/concepts/atomic/txn_mode_mvcc/
+    if check_mvcc_state() then
+        box.begin({txn_isolation = 'read-committed'})
+        max = self.space.index.task_id:max()
+        box.commit()
```
Totktonada, line 94: Can't we lean on some tarantool's facilities to autogenerate an identifier? For example:

* Use `sequence = true` index option and pass `nil` as an ID.
* Use uuid.

R-omk: Maybe just save counter in lua runtime.
Load from max() at init moment.

better0fdead: Yes, that is other options to fix that issue that we take into account. Currently we disscused to stick to the box.begin/commit.

queue/abstract/driver/utubettl.lua
better0fdead: discussed in a private call

Totktonada: Please, summarize it for ones who were not on your call.

queue/abstract/driver/fifo.lua
better0fdead: > Can you, please, reveal the reasons why the other alternatives were declined?
> 
> The given solution has the problem highlighted by @olegrok with nested transactions at least.

Sure. Problem highlighted by @olegrok  doesn't bother us because of 2 points:
1. Usecases when nested transactions will be opened are rare. (+ it will only happen when mvcc is on)
2. We didn't give promises that in that api there will be no box.begin calls. (that change will not be braking)
So we don't see that it is an issue.

Regarding  @R-omk  suggestion, it will need a lot of additional re-writing of 'queue' logic such as switch in  replicasets for example.

Given that @olegrok issue does'nt bother us. Current fix is enough.

queue/abstract/driver/fifo.lua
better0fdead: changed function

queue/abstract/driver/utubettl.lua
LeonidVas: Hi! The talk was discussing how an external transaction can occur and how valid it is, whether there have been use cases and how valid they are.
I have the following thoughts on this: the work is done with the top-level queue API `put()` (not with operations typical for a DB such as insert/select/...), which doesn't have any contracts on the internals, including the absence of transactions inside.

queue/abstract/driver/fifo.lua
LeonidVas: > Maybe just save counter in lua runtime. Load from max() at init moment.

Yes, we considered that option, but in this case we need to update that counter, when the master switched (replicaset mode on) and the current fix looks more localized.

LeonidVas: > Can't we lean on some tarantool's facilities to autogenerate an identifier? For example:
> 
> * Use `sequence = true` index option and pass `nil` as an ID.
> * Use uuid.

Switching to `uuid` will require alter several underlying spaces.
About using sequences, I need some time to think about whether that would work and be a better alternative.

LeonidVas (changes requested):

queue/abstract/driver/fifo.lua
```diff
@@ -66,9 +67,40 @@ function method.normalize_task(self, task)
     return task
 end
 
+-- check if mvcc is enabled
+-- returns true if mvcc is enabled
+local function check_mvcc_state()
+    if not check_version({2, 6, 1}) then
```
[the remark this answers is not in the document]
better0fdead: removed

queue/abstract/driver/fifo.lua
```diff
@@ -66,9 +67,40 @@ function method.normalize_task(self, task)
...
+
+    -- taking the maximum of the index is an implicit transactions, so it is
+    -- always done with 'read-confirmed' mvcc isolation level.
+    -- It can lead to errors when trying to make parallel 'put' calls with mvcc enabled.
+    -- It is hapenning because 'max' for several puts in parallel will be the same since
+    -- read confirmed isolation level makes visible all transactions that finished the commit.
+    -- To fix it we wrap it with box.begin/commit and set right isolation level.
+    -- See https://github.com/tarantool/queue/issues/207
+    -- See https://www.tarantool.io/ru/doc/latest/concepts/atomic/txn_mode_mvcc/
+    if check_mvcc_state() then
```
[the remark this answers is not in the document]
better0fdead: changed

queue/abstract/driver/fifo.lua
```diff
@@ -66,9 +67,40 @@ function method.normalize_task(self, task)
...
+
+    return false
+end
+
 -- put task in space
 function method.put(self, data, opts)
-    local max = self.space.index.task_id:max()
+    local max
+
+    -- taking the maximum of the index is an implicit transactions, so it is
```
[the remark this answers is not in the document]
better0fdead: changed

t/220-mvcc.t
```diff
@@ -0,0 +1,63 @@
...
+test:is(tube.name, 'test', 'tube.name')
+test:is(tube.type, 'fifo', 'tube.type')
+
+test:test('concurent take', function(test)
+    test:plan(151)
+
+    local channel = fiber.channel(1000)
+    test:ok(channel, 'channel created')
+
+    local res = {}
```
LeonidVas, line 36: What is this table for?

t/220-mvcc.t
```diff
@@ -0,0 +1,63 @@
...
+        fiber.create(function(i)
+            local taken = tube:take(1)
+            test:ok(taken, 'Task was taken ' .. i)
+            table.insert(res, { taken })
+            channel:put(true)
+        end, i)
+    end
+
+    fiber.sleep(.1)
+    test:ok(tube:put(1), 'task 1 was put')
```
LeonidVas, line 47: Why is this out of the loop?

t/220-mvcc.t
```diff
@@ -0,0 +1,63 @@
...
+    for i = 1, 50 do
+        fiber.create(function(i)
+            local taken = tube:take(1)
+            test:ok(taken, 'Task was taken ' .. i)
+            table.insert(res, { taken })
+            channel:put(true)
+        end, i)
+    end
+
+    fiber.sleep(.1)
```
LeonidVas, line 46: Why is there a need for `sleep()` here?

t/220-mvcc.t
```diff
@@ -0,0 +1,63 @@
...
+    test:ok(tube:put(1), 'task 1 was put')
+
+    for i = 2, 50 do
+        fiber.create(function(i)
+            test:ok(tube:put(i), 'task ' .. i .. ' was put')
+        end, i)
+    end
+    fiber.sleep(.1)
```
LeonidVas, line 54: Why is there a need for `sleep()` here?

t/220-mvcc.t
```diff
@@ -0,0 +1,63 @@
...
+test:test('concurent take', function(test)
+    test:plan(151)
+
+    local channel = fiber.channel(1000)
+    test:ok(channel, 'channel created')
+
+    local res = {}
+    for i = 1, 50 do
+        fiber.create(function(i)
+            local taken = tube:take(1)
```
LeonidVas, line 39: AFAIU `take()` is fixed in the next commit and should still be broken here.

t/220-mvcc.t
```diff
@@ -0,0 +1,63 @@
...
+    test:ok(tube:put(1), 'task 1 was put')
+
+    for i = 2, 50 do
+        fiber.create(function(i)
+            test:ok(tube:put(i), 'task ' .. i .. ' was put')
+        end, i)
+    end
+    fiber.sleep(.1)
+    for i = 1, 50 do
+        test:ok(channel:get(1 / i), 'take was done ' .. i)
```
LeonidVas, line 56: Do you use `channel()` to avoid `fiber_obj:join()`?

t/220-mvcc.t
```diff
@@ -0,0 +1,63 @@
...
+
+test:ok(rawget(box, 'space'), 'box started')
+test:ok(queue, 'queue is loaded')
+
+local tube = queue.create_tube('test', 'fifo', { engine = engine })
+test:ok(tube, 'test tube created')
+test:is(tube.name, 'test', 'tube.name')
+test:is(tube.type, 'fifo', 'tube.type')
+
+test:test('concurent take', function(test)
```
LeonidVas, line 30: Add a comment describing what the test checks.

LeonidVas (changes requested):
Hi! Thank you for the patchset.
See several comments bellow.

t/220-mvcc.t
better0fdead: removed

t/220-mvcc.t
better0fdead: removed

t/220-mvcc.t
better0fdead: removed

t/220-mvcc.t
better0fdead: changed

t/220-mvcc.t
better0fdead: removed

t/220-mvcc.t
better0fdead: added

LeonidVas: No)

t/220-mvcc.t
LeonidVas: No

t/220-mvcc.t
better0fdead: added

t/220-mvcc.t
better0fdead: Changed. Use channels to check results and to wait for fibres to finish work.

t/220-mvcc.t
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env tarantool
+local qc = require('queue.compat')
+local log   = require('log')
+if not qc.check_version({2, 6, 1}) then
+    log.info('Tests skipped, tarantool version < 2.6.1')
+    return
+end
+local yaml  = require('yaml')
```
LeonidVas, line 8: As far as I can see, you are not using yaml below.

LeonidVas (changes requested):

t/220-mvcc.t
better0fdead: deleted

PR #212, opened 2023-07-31
cd: add missing file to install target

0x501D:
`version.lua` is missing in RPM package.
packpack packs the result of executing `make install` on the target.


oleg-jukovec (approved):
Thank you for the patch!

PR #217, opened 2023-09-04
replicaset_mode: allow api call in init state

### LINKED ISSUE #216: in_replicaset mode: create_tube on replica problem:
Repro:

replica:
```lua
box.cfg{
  replication = ... 
  read_only = true
	---
}

queue = require("queue")
queue.cfg({ttr = 300, in_replicaset=true})
queue.create_tube('test_tube', 'fifottl', {temporary = false, if_not_exists = true})
```

Will fails with:
```
2023-09-04 14:01:11.000 [5990] main/103/replica.lua/box.load_cfg I> set 'read_only' configuration option to true
stack traceback:
	/home/void/tmp/cluster/repl/queue/init.lua:44: in function '__index'
	replica.lua:13: in main chunk
2023-09-04 14:01:11.004 [5990] main/105/checkpoint_daemon I> scheduled next checkpoint for Mon Sep  4 15:11:32 2023
2023-09-04 14:01:11.004 [5990] main utils.c:610 E> LuajitError: /home/void/tmp/cluster/repl/queue/init.lua:45: Please configure box.cfg{} in read/write mode first
```

`create_tube` method must correctly handle the situation of running a queue on a replica in replicaset mode.

0x501D:
In replicaset mode, the user can attempt to call public methods on the replica start.
For example, a single script is used for master and replica.
    
This patch allows you to call a public method in the INIT state without
interrupting the script by exiting with an obscure error.
It will only log an attempt to call a method from the check_state function,
which is present at the beginning of each public method.
    
Closes #216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,27 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+- In replicaset mode, the behavior of the public API is reduced to the same behavior
+  in all queue states, including INIT. Previously, in the INIT state, an ambiguous
+  error was thrown when trying to access a public method on a replica and the script
+  was interrupted by an error.
+
+  Old behavior (call `create_tube` on replica, queue is in INIT state):
+  ```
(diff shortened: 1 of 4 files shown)


0x501D:
On the other hand, you can close this pr and issue. Since this is not a bug, but a special case when the user, without checking the current state of the queue with the public `state()` method, calls the public method, and the queue at that moment is in the `INIT` state. In this case, the user script will abort with an ambiguous error.

LeonidVas (commented):
Hi! Thank you for the patch.

LeonidVas (approved):

PR #224, opened 2024-01-29
abstract: make fibers cancellable

nshy:
We are going to finish all client (non-system) fibers in the process of Tarantool shutdown. First we cancel them and then wait for their finishing. So if the client fibers are not finished Tarantool shutdown is never finished too.

Currently some of queue fibers can not be finished and we got hang on integration testing of PR [1]. Let's fix it.

[1] PR with client fibers finishing on shutdown.
https://github.com/tarantool/tarantool/pull/9604

### DIFF
--- a/queue/abstract/driver/utubettl.lua
+++ b/queue/abstract/driver/utubettl.lua
@@ -169,7 +169,7 @@ local function utubettl_fiber(self)
         if box.info.ro == false then
             local stat, err = pcall(utubettl_fiber_iteration, self, processed)
 
-            if not stat and not err.code == box.error.READONLY then
+            if not stat and not (err.code == box.error.READONLY) then
                 log.error("error catched: %s", tostring(err))
                 log.error("exiting fiber '%s'", fiber.name())
                 return 1
--- a/queue/abstract/queue_state.lua
+++ b/queue/abstract/queue_state.lua
@@ -85,8 +85,14 @@ local function create_state_fiber(on_state_change_cb)
                     log.info('Queue state changed: WAITING')
                 end
             end
+            -- Handle server shutdown.
+            if not pcall(fiber.testcancel) then
+                break
+            end
         end
     end)
+
+    log.info('Finished queue state fiber')
 end
 
 -- Public methods.


queue/abstract/driver/utubettl.lua
```diff
@@ -169,7 +169,7 @@ local function utubettl_fiber(self)
         if box.info.ro == false then
             local stat, err = pcall(utubettl_fiber_iteration, self, processed)
 
-            if not stat and not err.code == box.error.READONLY then
+            if not stat and not (err.code == box.error.READONLY) then
```
LeonidVas, line 172: Please move this to a separate commit.

queue/abstract/queue_state.lua
```diff
@@ -65,25 +65,21 @@ local function create_state_fiber(on_state_change_cb)
...
-                if rc then
-                    current = queue_state.states.STARTUP
-                    log.info('Queue state changed: STARTUP')
-                    -- Wait for maximum upstream lag * 2.
-                    fiber.sleep(max_lag() * 2)
-                    on_state_change_cb(current)
-                    current = queue_state.states.RUNNING
-                    log.info('Queue state changed: RUNNING')
-                end
+                box.ctl.wait_rw()
```
LeonidVas, line 68: If I understand correctly, we can keep calling with `pcall` and in case of an error write a message to the log, check if `fiber` is canceled and then throw a new error with more additional context. If this is the case, I prefer to do the extra processing.

LeonidVas (changes requested):
Hi! Thank you for the patch.
See several comments bellow.

queue/abstract/driver/utubettl.lua
nshy: Done.

queue/abstract/queue_state.lua
nshy: Changed to:

```diff
--- a/queue/abstract/queue_state.lua
+++ b/queue/abstract/queue_state.lua
@@ -85,8 +85,12 @@ local function create_state_fiber(on_state_change_cb)
                     log.info('Queue state changed: WAITING')
                 end
             end
+            -- Handle server shutdown.
+            if not pcall(fiber.testcancel) then break end
         end
     end)
+
+    log.info('Finished queue state fiber')
 end

 -- Public methods.
```

LeonidVas (approved):

PR #248, opened 2025-08-13
core: disable rw triggers on replicas

ShelbyKS:


Col-Waltz:
Thank you for the patch!

Please provide proper commit message for your changes:

```
core: disable rw triggers on replicas

<Some information about changes and implementation reason>
```

ShelbyKS:
provided propper commit message

Col-Waltz:
Thanks for providing a proper commit message, now I see some problems in this solution:

1) ERR_READONLY error signals that we try to modify data on RO instance. It is more safe to rely on `box.info.ro` to protect RO instanse from modifying.

2) `_on_consumer_disconnect` wakes all waiters so they can close their connections. Waiters operate with local fields and do not modify data itself, but it is useful if this instance will become master again. So it is safe to wake them up and it is better to do this so that they don't wait ttl.

I suggest to fix `_on_consumer_disconnect` like this:

```
local consumers = box.space._queue_consumers
local waiters = consumers.index.pk:select{conn_id}

for _, waiter in ipairs(waiters) do
        local fid = waiter[2]
        local cond = conds[fid]
        if cond then
            releasing_connections[fid] = true
            cond:signal(fid)
        end

        if not box.info.ro then
            consumers:delete{waiter[1], waiter[2]}
        end
    end
```

So we will wake up waiters, but delete them only if it is safe.

And `session_disconnect` also should be protected:

```
if not box.info.ro then
    session.disconnect(conn_id)
end
```

I will implement this changes myself in other PR.

Col-Waltz:
Implemented via https://github.com/tarantool/queue/pull/252

PR #249, opened 2025-08-14
NOTASK: teach queue grant tubes for roles

ShelbyKS:
There is ability to make tube grants just for users. Also need for roles


Col-Waltz:
Thank you for the patch!

Please provide proper commit message for your changes:

```
core: provide roles ability to grant tubes

<Some information about changes and implementation reason>
```

Col-Waltz:
Implemented via https://github.com/tarantool/queue/pull/253

PR #252, opened 2026-06-02
core: protect ro nodes on disconnect

Col-Waltz:
Adds protection for on_disconnect trigger. After leader change it may return ERR_READONLY while deleting consumers. Now this trigger performs only local operations on ro node.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- `ERR_READONLY` errors on previous leader after leader change in cluster
+   while processing `on_disconnect` triggers (#248).
+
 ## [1.4.4] - 2025-05-26
 
 The patch release fixes incorrect behavior of the utubettl driver with enabled
--- a/queue/abstract.lua
+++ b/queue/abstract.lua
@@ -479,6 +479,10 @@ end
 
 --- Release all session tasks.
 local function release_session_tasks(session_uuid)
+    if box.info.ro then
+        return
+    end
+
     local taken_tasks = box.space._queue_taken_2.index.uuid:select{session_uuid}
 
     for _, task in pairs(taken_tasks) do
@@ -501,26 +505,25 @@ end
 
 function method._on_consumer_disconnect()
     local conn_id = connection.id()
+    local consumers = box.space._queue_consumers
 
     -- wakeup all waiters
-    while true do
-        local waiter = box.space._queue_consumers.index.pk:min{conn_id}
-        if waiter == nil then
-            break
-        end
-        -- Don't touch the other consumers
-        if waiter[1] ~= conn_id then
-            break
-        end
-        box.space._queue_consumers:delete{waiter[1], waiter[2]}
-        local cond = conds[waiter[2]]
+    for _, waiter in consumers.index.pk:pairs(conn_id, { iterator = 'EQ' }) do
+        local fid = waiter[2]
+        local cond = conds[fid]
         if cond then
-            releasing_connections[waiter[2]] = true
-            cond:signal(waiter[2])
+            releasing_connections[fid] = true
+            cond:signal(fid)
+        end
+
+        if not box.info.ro then
+            consumers:delete{waiter[1], waiter[2]}
         end
     end
 
-    session.disconnect(conn_id)
+    if not box.info.ro then
+        session.disconnect(conn_id)
+    end
 end
 
 -- function takes tuples and recreates tube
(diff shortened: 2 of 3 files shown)


queue/abstract.lua
```diff
@@ -479,6 +479,10 @@ end
 
 --- Release all session tasks.
 local function release_session_tasks(session_uuid)
+    if box.info.ro then
+        return
+    end
```
oleg-jukovec, line 484: What happens in this case:

1. Instance -> RO.
2. Disconnect the session.
3. Instance -> RW.

?

It looks like the session will still be in the space.

queue/abstract.lua
```diff
@@ -501,26 +505,25 @@ end
 
 function method._on_consumer_disconnect()
```
oleg-jukovec, line 506: The same here.

oleg-jukovec (changes requested):
Please, add a changelog entry + link to the PR: https://github.com/tarantool/queue/pull/248

queue/abstract.lua
Col-Waltz: This callback operates with field `_queue_taken_2` which contains information about taken tasks. This field is **not temporary** if `in_replicaset==true` . On new rw instance inside `on_state_change` function `tube_release_all_orphaned_tasks` cleans all `TAKEN` tasks if they do not have live shared session.

Since this field is replicating after new rw instance started on previous instance this field will be updated by standard replication mechanism.

queue/abstract.lua
Col-Waltz: If this called after instance became ro all waited fibers will be awakened. They later will stop automatically because all connections to ro node will break.

On ro instance `_queue_consumers` and `_queue_session_ids` is not cleared. This fields are always temporary and will be cleaned and recreated if this instance will be rw again.

`session.disconnect` also performs DML and is protected from execution on ro. 

`session.disconnect` also modifies _queue_shared_sessions and sets current session `active=false` and `ttl`. This operation is prohibited on ro nodes but is not dangerous because new master sets all sessions inactive on startup. For more context look in `queue/abstract/queue_session.lua` on function `identification_init`. It is always called on new master in `on_state_change(STARTUP)` by `session.start()`.

t/240-ro-on-disconnect.t
```diff
@@ -0,0 +1,136 @@
...
+    test:ok(task ~= nil, 'task taken')
+
+    local taken_before = box.space._queue_taken_2.index.uuid:select{session_uuid}
+    test:is(#taken_before, 1, '_queue_taken_2 has 1 record before')
+
+    -- Call on_session_remove callback directly on RW: must release task.
+    local ok_call, err = pcall(session._on_session_remove, session_uuid)
+    test:ok(ok_call, ('on_session_remove ok on RW, err=%s'):format(tostring(err)))
+
+    -- taken record must be removed
```
oleg-jukovec:
```suggestion
    -- Taken record must be removed.
```

t/240-ro-on-disconnect.t
```diff
@@ -0,0 +1,136 @@
...
+    local taken_after = box.space._queue_taken_2.index.uuid:select{session_uuid}
+    test:is(#taken_after, 0, '_queue_taken_2 record removed')
+
+    -- task must become READY again and be takeable
```
oleg-jukovec:
```suggestion
    -- Task must become READY again and be takeable.
```

t/240-ro-on-disconnect.t
```diff
@@ -0,0 +1,136 @@
...
+    test:ok(task2 ~= nil, 'task is takeable again after release')
+    test:is(task2[3], 'data2', 'task data preserved')
+
+    client:close()
+end)
+
+rawset(_G, 'queue', nil)
+tnt.finish()
+os.exit(test:check() and 0 or 1)
+-- vim: set ft=lua :
```
oleg-jukovec, line 192: Do we really need the line for vim?

oleg-jukovec (changes requested):
Please, describe the change for a user in the CHANGELOG.md.

t/240-ro-on-disconnect.t
Col-Waltz: I'm not sure is it needed or not but I made this for consistency with other tests in `t/`

t/240-ro-on-disconnect.t
Col-Waltz: Fixed, thanks.

t/240-ro-on-disconnect.t
Col-Waltz: Fixed, thanks.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Unprotected space delete by `on_disconnect` trigger on read-only instance after
+  leader change (#248).
```
oleg-jukovec: The changelog should be from a use point of view, not a developer.

oleg-jukovec (changes requested):

CHANGELOG.md
Col-Waltz: Updated the changelog.

oleg-jukovec (approved):

queue/abstract.lua
```diff
@@ -539,10 +542,29 @@ local function recreate_tube(tube_tuple)
     return make_self(driver, space, name, tube_type, id, opts)
 end
 
+-- function cleans local temporary spaces on startup to avoid
```
palage4a: A sentence should be started with capital letter.
https://www.tarantool.io/en/doc/latest/contributing/lua_style_guide/#commenting

```suggestion
-- Cleans local temporary spaces on startup to avoid
```

queue/abstract.lua
```diff
@@ -539,10 +542,29 @@ local function recreate_tube(tube_tuple)
...
+    if s ~= nil then
+        s:truncate()
+    end
+end
+
 -- Function takes new queue state.
 -- The "RUNNING" and "WAITING" states do not require additional actions.
 local function on_state_change(state)
     if state == queue_state.states.STARTUP then
+        cleanup_temp_spaces()
```
palage4a, line 567: Could you please provide the test that checks cleanup after RW->RO->RW transition? 
It seems to me, this case does not covered.

palage4a (changes requested):

queue/abstract.lua
Col-Waltz: Thanks, fixed.

queue/abstract.lua
Col-Waltz: Thanks, I provided a test case for this.

palage4a (approved):
LGTM

Issue #231, opened 2024-05-14
driver: possible data races in `kick` method


DerekBum:
`kick` method is used to unbury several tasks in the tube. After analyzing this method for drivers `fifo`, `fifottl`, `utube`, `utubettl`, I came to the conclusion, that it still possible to get a data race and incorrect, unexpected result.

https://github.com/tarantool/queue/blob/aa7c092a1f2ac80d3cfc63c75595d10c5a95c908/queue/abstract/driver/fifo.lua#L158-L173

We take a task with minimal `task_id` and `BURRIED` status. After that we are updating it, commiting. But lua can also yield before WAL write. So this very task could also be taken in a parallel `kick` call for the second time.

Possible fix: do changes inside a transaction, like in `put`, `take` methods. But there is a subtle point here (which is described in a comment below).
https://github.com/tarantool/queue/blob/aa7c092a1f2ac80d3cfc63c75595d10c5a95c908/queue/abstract/driver/fifo.lua#L101-L117

```[tasklist]
### Tasks
- [ ] `fifo`
- [ ] `fifottl`
- [ ] `utube`
- [ ] `utubettl`
```

PR #258, opened 2026-07-31
drivers: fix utube kick in the ready buffer mode

### LINKED ISSUE #256: utube: kick() corrupts the ready buffer in the ready_buffer storage mode
Two defects in `method.kick()` of the `utube` driver, both only in the `STORAGE_MODE_READY_BUFFER` mode. `utubettl` does the same thing correctly via its `update_ready()` helper.

https://github.com/tarantool/queue/blob/03cece7679b6fb6b6f94b90af2842db91cc2cdf8/queue/abstract/driver/utube.lua#L422-L433

**1. Wrong argument list.** When the utube has no entry in the ready buffer, `kick()` calls `put_ready(self, task[3])`, while the helper is `put_ready(self, id, utube)`. The utube name arrives as the task id and the utube becomes `nil`, so the insert fails on the space format and the error is swallowed by the `pcall` inside `put_ready()`. The kicked task stays READY but is never returned by `take()`, because `take_ready()` only looks at the ready buffer.

```lua
local tube = queue.create_tube('t', 'utube',
    {storage_mode = queue.driver.utube.STORAGE_MODE_READY_BUFFER})
local task = tube:put('single', {utube = 'u'})
tube:bury(task[1])
tube:kick(1)                     -- returns 1
box.space.t_ready_buffer:select()  -- [] , expected [[0, 'u']]
tube:take(0)                     -- nil, the task is stuck
```

**2. Status written instead of the utube name.** When the utube already has a younger task in the buffer, `kick()` replaces it with `insert({task[1], task[2]})`, that is `{id, status}` instead of `{id, utube}`. The status string lands in the utube field, so the buffer no longer has an entry under the real utube name and a later `put()` adds a second one. With two entries for one utube the driver can take two tasks of the same utube at once, which breaks the main utube guarantee, and `take_ready()` can spin forever on an entry whose utube already has a TAKEN task.

**3. Leaked transaction.** Both early returns of `kick()` return without calling `commit_func()`, so with no buried tasks the transaction opened by `begin_if_not_in_txn()` stays running and leaks into whatever the caller fiber does next:

```lua
tube:kick(1)      -- nothing is buried
box.is_in_txn()   -- true
```

`utubettl` commits before both early returns. This part is within the scope of #231, the rest is not.

bigbes:
Two defects in method.kick() of the utube driver, both in the 'ready_buffer' storage mode. utubettl does the same thing correctly via update_ready().

1. When the utube had no entry in the ready buffer, kick() called `put_ready(self, task[3])` while the helper is `put_ready(self, id, utube)`. The utube name was passed as the task id and the utube became nil, so the insert failed on the space format and the error was swallowed by the pcall inside put_ready(). The kicked task stayed READY but was never returned by take(), because take_ready() only looks at the ready buffer.

2. When the utube already had a younger task in the buffer, kick() replaced it with `insert({task[1], task[2]})`, that is {id, status} instead of {id, utube}. The status string landed in the utube field, so the buffer no longer had an entry under the real utube name and a later put() added a second one. With two entries for one utube the driver could take two tasks of the same utube at once, which breaks the main utube guarantee, and take_ready() could spin forever on an entry whose utube already had a TAKEN task.

Introduce update_ready() mirroring the utubettl helper and use it in kick().

Also commit the transaction before the early returns of kick(): with no buried tasks kick() returned while the transaction opened by begin_if_not_in_txn() was still running, and it leaked into whatever the caller fiber did next. This part is within the scope of #231.

Closes #256


t/260-utube-kick.t
```diff
@@ -0,0 +1,86 @@
...
+        tube:truncate()
+    end
+
+    -- A kicked task that displaces a younger task of the same utube must be
+    -- written to the buffer with the utube name, not with its status.
+    local first  = tube:put('first',  {utube = 'v'})
+    local second = tube:put('second', {utube = 'v'})
+    tube:take(0)             -- takes 'first', drops it from the buffer
+    tube:bury(first[1])      -- buries it, 'second' takes its buffer slot
+    tube:kick(1)             -- kicks 'first' back, it displaces 'second'
```
oleg-jukovec:
```suggestion
    tube:take(0)             -- Takes 'first', drops it from the buffer.
    tube:bury(first[1])      -- Buries it, 'second' takes its buffer slot.
    tube:kick(1)             -- Kicks 'first' back, it displaces 'second'.
```

t/260-utube-kick.t
```diff
@@ -0,0 +1,86 @@
...
+    test:is_deeply(rows, {{first[1], 'v'}},
+        'the displacing task keeps the utube name in the buffer')
+end)
+
+test:test('kick does not leak a transaction', function(test)
+    test:plan(1)
+
+    local tube = queue.create_tube('kick_txn', 'utube', {engine = engine})
+
+    tube:kick(1)  -- nothing is buried, kick returns early
```
oleg-jukovec:
```suggestion
    tube:kick(1)  -- Nothing is buried, kick returns early.
```

oleg-jukovec (approved):
Thank you for the patch. Please, fix comments.

t/260-utube-kick.t
bigbes: Fixed, ty!

t/260-utube-kick.t
bigbes: Fixed, ty!

Col-Waltz (approved):

PR #259, opened 2026-08-21
Extend driver API and add subqueuettl driver

ThCompiler:
Extend driver API:

Added the optional consumer_group driver API so a driver can wake a consumer waiting for the same logical group instead of waking an unrelated consumer.

Added custom statistics for drivers to support showing custom statistics by them.

Subqueuettl: 

Some workloads need to create queues dynamically with arbitrary names. Creating a separate tube for each queue requires converting those names to valid Tarantool space names and granting access whenever a new tube is created. This adds operational overhead and makes access management harder.

The utube model is otherwise well suited for this use case: it stores multiple logical queues in one space and guarantees sequential processing inside each queue. However, it does not support taking a task from a specific utube, so it cannot be used when consumers must be assigned to particular dynamically named queues.

Added subqueuettl as a TTL-aware alternative. It stores multiple dynamically named subqueues in one space, and both put() and take() require a subqueue name. This keeps task selection isolated while retaining TTL, TTR, priority, delay, and ready-buffer support.

If you don’t think it’s necessary to add this driver to the general library, I can remove it from this PR, but for my purposes, I need support for changes in the “abstract.lua” in order to use this driver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,7 +647,7 @@ or it may be acted on by a worker (usually with a `take` request).
 ## Taking a task from the queue ("consuming")
 
 ```lua
-queue.tube.tube_name:take([timeout])
+queue.tube.tube_name:take([timeout [, {options} ]])
 ```
 
 Take a queue task.
--- a/README.md
+++ b/README.md
@@ -825,7 +825,11 @@ queue.statistics( [queue name] )
 
 Show the number of tasks in a queue broken down by `task_state`, and the number
 of requests broken down by the type of request. If the queue name is not
-specified, show these numbers for all queues.
+specified, show these numbers for all queues. 
+
+In addition, any driver can add extra information for statistics by implementing their own `statistics` method.  
+The result of this method will be added as additional key of the returned value `queue.statistics`.
+
 Statistics are temporary, they are reset whenever the Tarantool server restarts.
 
 Example:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-### Changed
+- Allow pass options to `take()` via driver API.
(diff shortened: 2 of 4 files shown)


queue/abstract/driver/subqueuettl.lua
```diff
@@ -0,0 +1,608 @@
...
+            return self.space:update(id, {
+                { '=', i_status, state.DELAYED },
+                { '=', i_next_event, util.event_time(opts.delay) },
+                { '+', i_ttl, util.time(opts.delay) }
+            })
+        end
+
+        return self.space:update(id, {
+            { '=', i_status, state.READY },
+            { '=', i_next_event, util.time(task[i_created] + task[i_ttl]) }
```
Col-Waltz: Why do you call `util.time()` which convert seconds to microseconds on `i_ttl` and `i_created` which are already in microseconds?

README.md
```diff
@@ -302,6 +303,41 @@ seconds; if `ttr` is not specified, it is set to the same as `ttl`
...
+local tube = queue.create_tube('sites', 'subqueuettl')
+tube:put('https://example.com', {subqueue = 'example.com'})
+local task = tube:take(10, {subqueue = 'example.com'})
+```
+
+`subqueuettl` adds per-subqueue statistics under the common `driver` field:
+
+```lua
+local stats = queue.statistics('sites')
+local example_stats = stats.driver.subqueues['example.com']
```
Col-Waltz: I suppose there is a typo because in `t/250-subqueuettl.t` and later in `README` you use `stats.extra.subqueues.stat_1.taken` to get stats.

Col-Waltz (changes requested):

README.md
ThCompiler: fixed

queue/abstract/driver/subqueuettl.lua
ThCompiler: removed `util.time()`

Col-Waltz (approved):

